PR #8, opened 2022-07-29
7 create cd pipeline

### LINKED ISSUE #7: Create CD pipeline
Use github actions to create a CD pipeline to lint python code and run unit tests

simonrnss:
Closes #7
Created a CD pipeline to lint using `pylint` and run tests using `unittest`.

Currently, setup as:
1. `pylint` passes if code is rated at least 7. Happy to change this. We can also ask pylint to ignore some things if we don't care about them.
2. `unittest` is run to discover any tests in the `tests` folder. Happy to change this also.

### DIFF
--- a/.github/workflows/lint-test-on-push.yml
+++ b/.github/workflows/lint-test-on-push.yml
@@ -1,4 +1,4 @@
-name: Python package
+name: pylint and unit test
 
 on: [push]
 


rpreen:
We should probably add the contents of the tests folder in this PR so that there are some tests to pickup and check that this works?

The tests for safemodel were based on pytest and I think that running `pytest .` should pickup both pytests and unittests.

simonrnss:
> We should probably add the contents of the tests folder in this PR so that there are some tests to pickup and check that this works?
> 
> The tests for safemodel were based on pytest and I think that running `pytest .` should pickup both pytests and unittests.

Good plan -- could you add please? It's def working for unittest as my other branches come from this branch and the tests are running there.

Will `pytest` need to be added to the dependencies in requirements.txt?

rpreen:
It looks like most of the tests require the `attacks` and `attack_utilities` imports so we need to add those first.

rpreen:
Since it looks like the next PR modifies this one maybe these should just get merged and then we can add the attack utilities and the rest of the tests?

rpreen (approved):

PR #20, opened 2022-08-03
add more tests

rpreen:
Currently adds (and passes tests):
* attack_utilities/
* tests/test_safedecisiontreeclassifier.py
* tests/test_saferandomforestclassifier.py

Omitted due to (missing) data requirement:
* tests/test_safesvc.py

Do we need to add data fetching to the CI runner to run such tests?


attack_utilities/estimator_template.py
```diff
@@ -0,0 +1,86 @@
+'''
```
simonrnss, line 1: I think this probably needs better commenting / docs.
Do we actually use it though? I thought these classes had been superseded by the safe wrappers?

attack_utilities/dp_svc_example.py
```diff
@@ -0,0 +1,113 @@
+'''
```
simonrnss, line 1: Not sure we need this code? And if we do, it would find a more natural home in the examples folder?

simonrnss (changes requested):
A couple of suggestions -- not sure why we need the attack_utilities code? Where is it being imported?

simonrnss:
@rpreen please also check pylint output -- if you're happy to overrule its variable name issues then adding
```python
# pylint: disable = invalid-name
```
to the offending files will shut pylint up. It's also finding near duplicate code in the test files. This might be unavoidable.

rpreen:
Yeah, the `attack_utilities` should be removed if it's not needed. I think the pylint clean up should be a new issue/PR though since most of the complaints are about the safe model modules which aren't really dealt with in this PR (this is about adding tests)

mccarthyajb:
OK, to be clear that's removing the entire attack_utilities directory? 
I don't think safesvc references it anymore.

rpreen:
yes

mccarthyajb:
OK, we were still importing GenericEstimator and the DPSVC class was inheriting it. 
I've removed those dependencies and the attack_utilities directory. The test are successful without them.

PR #21, opened 2022-08-10
12 add likelihood test example

### LINKED ISSUE #12: Add likelihood test example
Create an example script for the likelihood test showing how to call it programmatically and from the command line

simonrnss:
Closes #12 

Adds an example of running the lira attack, showing how to call the code programmatically, and from the command line.

Also

- renamed likelihood_test to likelihood_attack
- refactored AUC p value to move the code out of the example

### DIFF
--- a/.github/workflows/lint-test-on-push.yml
+++ b/.github/workflows/lint-test-on-push.yml
@@ -1,3 +1,4 @@
+---
 name: pylint and unit test
 
 on: [push]
@@ -27,3 +28,4 @@ jobs:
       - name: Test with pytest
         run: |
           pytest .
+...


rpreen:
I don't think github seems to particularly care but yaml files are supposed to start with a single line with `---` and end with a single line with `...`. See: [specification](https://yaml.org/spec/1.2.2/#22-structures)

simonrnss:
> I don't think github seems to particularly care but yaml files are supposed to start with a single line with `---` and end with a single line with `...`. See: [specification](https://yaml.org/spec/1.2.2/#22-structures)

Thanks -- will amend

rpreen (approved):

PR #29, opened 2022-08-17
22 make metrics methods private

### LINKED ISSUE #22: Make metrics methods private
Do a parse through `attacks/metrics.py`, ensuring all methods that would (and should) not be called directly are private (i.e. name starts with an underscore). This will automatically stop them appearing in the documentation.

Methods to change:
- `metrics.div`
- `metrics.expected_auc_var`
- `metrics.tpr_at_fpr`

simonrnss:
Closes #22 

Methods made private and not called within codebase.


mccarthyajb (commented):
Great, you've made div, tpr_at_fpr, and expected_auc_var private. You're also now using PDF01 instead of PDIF. I note there's a unused import showing for attacks_worst_case. Is this because we're not using the scipy stats stuff anymore?

simonrnss:
Thanks -- removed that line

PR #38, opened 2022-08-26
Example notebooks

mccarthyajb:
Hi @jim-smith can you please review the Example notebooks. They're linked from the sphinx documentation in docs/source/examples.rst


jim-smith:
@mccarthyajb Andy, 
1. it would probably be more useful to readers if we saved these with the outputs present.
2. In the DT example, possibly all of them, the first cell adds some things to the sys.path variables.     That probably needs a quick md cell above  explaining, especially as the second set of things you add is quite dependent on where things are on your machine 
3. README.md needs expanding - for instance say that the dT one is the main one to read because it talks through the different aspects of the functionality added by the safe wrappers.
4. in the safeDT one, please would you add a few md cells to give it structure and explain what you are doing when you call the posthoc_hoc check and request release() functions.

jim-smith:
good to go!

PR #39, opened 2022-08-30
37 howto more models

mccarthyajb:
Hi @jim-smith, I think this is ready to merge. 

The changes will only become apparent in the sphinx docs when they're merged to main. 

The template in safemodel/classifiers has place-holder text "ModelToMakeSafer" that I imagine will be globally searched and replaced. 

Thanks, 

Andy.

### DIFF
--- a/docs/source/extending.rst
+++ b/docs/source/extending.rst
@@ -168,6 +168,3 @@ safekeras.py that help with the the specifics of neural networks.
             if not np.array_equal(m1d, m2d):
                 return False, f"dimension {dim} of layer {layer} differs"
 	    return True, "weights match"
-
-
-
--- a/docs/source/extending.rst
+++ b/docs/source/extending.rst
@@ -25,9 +25,9 @@ Copy The Template
 
 .. code-block:: shell
 
-	{
+	
 		cp new_model_template xgboost.py
-	}
+	
 
 ::
 
@@ -36,11 +36,11 @@ Define the Safer Class
 
 .. code-block:: python
 
-	{
+	
 	class SafeGradientBoosting(SafeModel, GradientBoostingClassifier):
 		"""Privacy protected GradientBoostingClassifier."""
 
-	}
+	
 
 ::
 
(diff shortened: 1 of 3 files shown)


mccarthyajb:
pre-commit.ci autofix

jim-smith:
@mccarthyajb 
Hi Andy,
one small change please
in order to integrate running attacks I had to edit the __init__() methods to include a list of the attributes t hat are defined and count as valid params to be passed to the other class definition.

- The extra code goes between the calls the two superclass inti functions, and the params for the second one are changed

- **The only  bit that people will need to change for a new class is the contents of the list self.basemodel_paramnames***. 
  and the code that builds the_kwds should be generic
- I got the list for the ones I have done just by copy-pasting from the sklearn man pages

For example that method in SafeRandomForest now looks like 
```python=
    def __init__(self, **kwargs: Any) -> None:
        """Creates model and applies constraints to params"""
        SafeModel.__init__(self)
        self.basemodel_paramnames=[
            'n_estimators','criterion','max_depth','min_samples_split',
            'min_samples_leaf','min_weight_fraction_leaf','max_features',
            'max_leaf_nodes','min_impurity_decrease','bootstrap',
            'oob_score','n_jobs','random_state','verbose'
            'warm_start','class_weight','ccp_alpha','max_samples']

        the_kwds=dict()
        for key,val in kwargs.items():
            if key in self.basemodel_paramnames:
                the_kwds[key]=val
        RandomForestClassifier.__init__(self, **the_kwds)
        self.model_type: str = "RandomForestClassifier"
        super().preliminary_check(apply_constraints=True, verbose=True)
        self.ignore_items = [
            "model_save_file",
            "ignore_items",
            "base_estimator_",
        ]
        self.examine_seperately_items = ["base_estimator", "estimators_"]
```

jim-smith:
@mccarthyajb I've edited the template file to reflect the code that needs adding/changing

mccarthyajb:
Thanks @jim-smith, I've updated the __init__ section of the .rst as here:

https://github.com/AI-SDC/AI-SDC/blob/37-HOWTO-more-models/docs/source/extending.rst

jim-smith:
@mccarthyajb completed

PR #40, opened 2022-08-31
33 standard reporting strings

mccarthyajb:


mccarthyajb:
Hi @jim-smith, This code works and I have manually run the pytests.
(base) andy@ubuntu:~/AI-SDC$ pytest tests/
========================================================= test session starts ==========================================================
platform linux -- Python 3.9.7, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
rootdir: /home/andy/AI-SDC
plugins: anyio-2.2.0
collected 36 items                                                                                                                     

tests/test_data_interface.py ....                                                                                                [ 11%]
tests/test_metrics.py ....                                                                                                       [ 22%]
tests/test_safedecisiontreeclassifier.py ........                                                                                [ 44%]
tests/test_safekeras.py ............                                                                                             [ 77%]
tests/test_saferandomforestclassifier.py .......                                                                                 [ 97%]
tests/test_safesvc.py .                                                                                                          [100%]

=========================================================== warnings summary ===========================================================
../anaconda3/lib/python3.9/site-packages/flatbuffers/compat.py:19
  /home/andy/anaconda3/lib/python3.9/site-packages/flatbuffers/compat.py:19: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
    import imp

-- Docs: https://docs.pytest.org/en/stable/warnings.html
==================================================== 36 passed, 1 warning in 49.10s ====================================================
The functionality is there; unfortunately a number of pylint issues remain blocking this PR. Is this needed for the release baseline?

jim-smith:
@mccarthyajb Andy,     toner f the main pylintr complaints seems to be around the number of local variables.
you also have a very long bit of code that initialises them to values ink wargs if present else dome constant 
I think the solution would be to create a dict with al those values (in string form) as keys, explicitly initialising them to fill/0 /whatever, then do a single loop over newdict.keys() that looks for the key ink wargs and swats the value if required.

Then in the f-strings you'd have to use {newdict[varname]} rather than {varname} 
But I'm guessing you might not have time to do that today ...

mccarthyajb:
Hi @jim-smith, 

I've refactored the reporting module to use the dictionary idea.
We now have a clean pylint again. The pytests are al passed successfully.  

This is now ready to merge. 

Thanks, 

Andy.

jim-smith:
@mccarthyajb 
Hi Andy,
looks good - only thing I can see is that in safemodel.py, check_min/Check_max, where you've implemented. the check that the value passed for a parameter is int or float.  If it fails that you call the get_message() method with no parameters which returns a message that is maybe not that helpful.    Possibly you could  call the message for wrong type of parameter value instead?

jim-smith (commented):
@mccarthyajb
Hi Andy,
looks good - only thing I can see is that in safemodel.py, check_min/Check_max, where you've implemented. the check that the value passed for a parameter is int or float. If it fails that you call the get_message() method with no parameters which returns a message that is maybe not that helpful. Possibly you could call the message for wrong type of parameter value instead?

jim-smith:
@mccarthyajb just realised that the 'other side' has not been updated I.e. the various purest files.

Is this a commit that hasn't happened or do they need to go on the to-do list?

mccarthyajb:
pre-commit.ci run

mccarthyajb:
Trimmed trailing spaces from README.md
pre-commit.ci run

mccarthyajb:
Hi @Jim, I've changed message returned in check_min and check_max to use the type different than recommended type message. I've also fixed some issues with trailing spaces in the README.md

The pytest files currently use hard-coded text strings to check the output. 
Not sure if this is a problem; if we want the pytests to use the reporting strings that will need adding to the to-do list.

Thanks, 

Andy.

PR #47, opened 2022-09-07
added notebooks and supporting files

jim-smith:
@simonrnss @drchriscole I think I havre managed to track down the right notebook from a branch in the graimatter repo and bring them into this branch on ai-sdc.

Can you have a quick look to confirm then if correct, do the merge and let Smarti know the url

thanks

Jim 

ps:these were all done when I was on leave so I'm not familiar with them


simonrnss:
I think these are right, but will await word from Smarti.
Assuming they are....a couple of suggestions for the pr
1. Not sure we should have .html files in the repo? The .rmd files (from which the html was generated) would be enough
2. should we restructure the folder structure a bit to distinguish different types of examples a bit better -- these examples are to help explain what he risks are. The others are to explain how the code works. So maybe we're missing a layer of abstraction?

simonrnss:
And TODO: seems to be a couple of spelling errors in the R ones being picked up by the pre commit (ignore the zillions in the html as I think we should remove the files)

simonrnss:
On reflection, I agree -- they don't naturally fit here.
Maybe we should just merge them in the GRAIMatter github then and link there rather than here?

jim-smith:
I thought about that, but GRAIMatter is not public, nor would I want it to be, so this seemed like the best-worst option.

I could put them somewhere on my UWE home page, or a research group page,  but then the notebooks wouldn't get rendered.

jim-smith:
Or I could make a public repo from my own GitHub if you would prefer?

simonrnss:
Ah, ok, fair enough. The only alternative would be making _another_ repo, and lets not do that. We'll put them in AISDC but signpost them carefully.
I've sent one final query to Smarti -- once she responds, I'll rename them a bit, modify the directory structure, add something to the readme to make clear what these things are

PR #54, opened 2022-09-29
fix sphinx build

rpreen:
Sphinx build requires python 3.10 to build successfully.

This needs to be in `main` to work.


jim-smith:
sorted.

PR #71, opened 2022-10-10
format with black and fix implicit str conversions

rpreen:


rpreen:
Cancelling this. The tests don't even work correctly in development, so no wonder they don't pass here either...

PR #90, opened 2022-12-22
added option to not sort

### LINKED ISSUE #89: Additional params to WorstCaseModel
Whilst working on assessing an actual model, discovered a limitation in parameters we expose in the `WorstCaseModel` class.

In the `run_attack_reps` method (https://github.com/AI-SDC/AI-SDC/blob/7c68ac07d2e9d863fc3700625f820922f7ae2816/aisdc/attacks/worst_case_attack.py#L131) we default to sorting the probabilities:
https://github.com/AI-SDC/AI-SDC/blob/7c68ac07d2e9d863fc3700625f820922f7ae2816/aisdc/attacks/worst_case_attack.py#L151-L152
This makes sense in general, but there are situations when it doesn't make sense to sort the output.

Suggested fix is to add a new parameter to `WorstCaseAttackArgs` called `sort_probs` with default value = True. This will also need to be added to the `attack_parser` arg parse in the `main` method. Help text can be "whether or not to sort the predictive probabilities prior to training attack models. Default again should be True.

The following lines should only be run if `sort_probs` is True:
https://github.com/AI-SDC/AI-SDC/blob/7c68ac07d2e9d863fc3700625f820922f7ae2816/aisdc/attacks/worst_case_attack.py#L151-L152

simonrnss:
Closes #89 

The standard behaviour in MI attacks is to use the _sorted_ predictive probabilities as features. I.e. feature one is the highest probability. This makes the attack easier and less nonlinear (in an ad hoc kind of way of thinking).

This won't always be relevant so I've added an option to not sort them.

Also refactored part of `WorstCaseAttack` to expose this operation for testing.

### DIFF
--- a/aisdc/attacks/worst_case_attack.py
+++ b/aisdc/attacks/worst_case_attack.py
@@ -129,7 +129,10 @@ class WorstCaseAttack(Attack):
                 self.dummy_attack_metrics += temp_metrics
         logger.info("Finished running attacks")
 
-    def _prepare_attack_data(self, train_preds: np.ndarray, test_preds: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
+    def _prepare_attack_data(
+        self,
+        train_preds: np.ndarray,
+        test_preds: np.ndarray) -> tuple[np.ndarray, np.ndarray]:
         """Prepare training data and labels for attack model
         Combines the train and test preds into a single numpy array (optionally) sorting each
         row to have the highest probabilities in the first column. Constructs a label array that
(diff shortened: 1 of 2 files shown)


simonrnss:
gah...need to fix linting

jim-smith (changes requested):
Changes to sorting look good but did you mean to include the sparra image?

simonrnss:
Ah yes, I'll get rid of that

jim-smith (approved):

PR #105, opened 2023-03-15
Development shahzad

shahzadmumtaz22:
Issue 103 - https://github.com/AI-SDC/AI-SDC/issues/103 and Issue 109 https://github.com/AI-SDC/AI-SDC/issues/109
Worst_case_attack scenario - Fast Fail Implementation
- worst_case_attack.py
1) Fail fast option based on AUC metric for a given threshold value (or a default threshold value) while running attack repetitions. 
2) Updating logging message when attack repetitions are run to demonstrate (1-10 of 10 instead of 0-9 of 10). 
3) When calling the attack function in cases where we apply dummy repetitions (i.e. "n_dummy_reps =1"), we can set the values for 'train_beta' and 'test_beta' arguments when running attacks in a programmatic way, but on a command line call to run_attack we don't have options (command line arguments) to set value for 'train_beta' and 'test_beta' so we end up relying on default values provided within "worst_case_attack.py" file. An example is given in examples.worst_case_attack_example.py
- worst_case_attack_example.py
4) Running a worst-case attack scenario with checking AUC calculated on each repetition with the given threshold, and if it is above the given threshold, all the following repetitions should be suspended. The file is updated to reflect both programmatic and command-line worst-case attack calls.
- dataset.py
5) self.n_samples = len(x_train) + len(y_train) - current line of code
I am wondering should this not be like 
self.n_samples = len(x_train) + len(x_test) - updated line of code


aisdc/attacks/worst_case_attack.py
```diff
@@ -218,7 +220,7 @@ def run_attack_reps(  # pylint: disable = too-many-locals
         )
 
         mia_metrics = []
-        for rep in range(self.args.n_reps):
+        for rep in range(1, self.args.n_reps + 1, 1):
```
simonrnss, line 223: I think we should change this -- if the issue is just in the output, just add 1 to the output.

aisdc/attacks/worst_case_attack.py
```diff
@@ -218,7 +220,7 @@ def run_attack_reps(  # pylint: disable = too-many-locals
...
             logger.info("Rep %d of %d", rep, self.args.n_reps)
```
simonrnss, line 224: i.e. here just change to 
`logger.info("Rep %d of %d", rep + 1, self.args.n_reps)`

simonrnss (changes requested):
Hi, thanks for the changes. A few comments in the file....mainly
1. I think we can make the failfast more general so that it could work with any metric.
2. I'd like to avoid modifying the looping logic to start at 1 (it's pretty non-python) -- if you'd like the logging output to go 1, 2, 3 instead of 0, 1, 2 then just add 1 in the call to logger

aisdc/attacks/worst_case_attack.py
```diff
@@ -24,6 +24,7 @@
 logging.basicConfig(level=logging.INFO)
 
 P_THRESH = 0.05
+AUC_FAIL_THRESH = 0.8
```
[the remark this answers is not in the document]
simonrnss: In fact, I'd also maybe consider making the object stateful so that it has its own counter and only fails once a certain condition has been met N times.

The rationale is that it would be better (IMO) to be checking the AUC p-value (significance of AUC values massively depends on the size of the dataset (an AUC of 0.75 is within 3SDs of 0.5 for a test dataset with 20 examples from each class). However, one p-value < (say) 0.05 in 100 reps is not a problem (it'll almost certainly disappear on multiple testing)., so perhaps stopping if it happens 10 times, etc makes more sense.

Similarly for the p-value that comes from the FDIF metric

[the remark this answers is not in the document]
shahzadmumtaz22: Dear Simon, 

Thanks for the feedback/reviewing this PR. We discussed a few more things in last week's catch-up. I agree that we should make the fail-fast option more generic. Given your feedback and the discussion we had last week, I will make changes. Once all that is done, I will update this PR.

aisdc/attacks/worst_case_attack.py
shahzadmumtaz22: Agreed - will make changes accordingly

aisdc/attacks/worst_case_attack.py
```diff
@@ -242,6 +244,9 @@ def run_attack_reps(  # pylint: disable = too-many-locals
                     mia_metrics[-1]["yeom_tpr"] - mia_metrics[-1]["yeom_fpr"]
                 )
 
+            if mia_metrics[rep - 1]["AUC"] >= self.args.auc_fail_thresh:
```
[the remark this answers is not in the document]
shahzadmumtaz22: We discussed this last week, and I will make changes accordingly.

aisdc/attacks/worst_case_attack.py
shahzadmumtaz22: Agreed - will update this.

aisdc/attacks/worst_case_attack.py
```diff
@@ -426,23 +468,79 @@ def _construct_metadata(self):
...
+            temp_dummy_attack_metrics = self.dummy_attack_metrics[exp_rep]
+            dummy_attack_metric_instances = {}
+            for rep in range(len(temp_dummy_attack_metrics)):
+                dummy_attack_metric_instances[
+                    "instance_" + str(rep + 1)
+                ] = temp_dummy_attack_metrics[rep]
+            temp = {}
+            temp["attack_instance_logger"] = dummy_attack_metric_instances
+            temp[
+                "attack_metric_failfast_summary"
```
simonrnss, line 515: Would it be neater to not sumamrise here? Why not put in all the failfast stuff and summarise later? Or is this just a description to help describe why the experiment stopped?

simonrnss (changes requested):
Thanks for all this -- really good. See individual comments, although the main one is that we need quite a lot more commenting for methods to make it clearer what they are doing.

aisdc/attacks/failfast.py
```diff
@@ -0,0 +1,71 @@
...
+class FailFast:  # pylint: disable=too-many-instance-attributes
+    def __init__(self, metric_name, metric_success_thresh, comp_type):
+        self.metric_name = metric_name
+        self.metric_success_thresh = metric_success_thresh
+        self.comp_type = comp_type
+        self.success_count = 0
+        self.fail_count = 0
+
+    def check_attack_success(self, metric_dict):
+        metric_value = metric_dict[self.metric_name]
```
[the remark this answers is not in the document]
shahzadmumtaz22: Done.

aisdc/attacks/worst_case_attack.py
```diff
@@ -220,8 +245,15 @@ def run_attack_reps(  # pylint: disable = too-many-locals
         )
 
         mia_metrics = []
-        for rep in range(1, self.args.n_reps + 1, 1):
-            logger.info("Rep %d of %d", rep, self.args.n_reps)
+
+        failfast_metric_summary = FailFast(
```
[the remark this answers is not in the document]
shahzadmumtaz22: Done

jim-smith:
@shahzadmumtaz22 if you pull the latest version of safemodels/classifers/safekeras.py into this branch it should fix the listing and test issues caused by tensorflow versions changing recently
- or change the two lines different manually

aisdc/attacks/worst_case_attack.py
```diff
@@ -53,9 +62,9 @@ def __init__(self, **kwargs):
             "min_samples_leaf": 10,
             "max_depth": 5,
         }
-        self.__dict__["attack_metric_success_name"] = "AUC"
-        self.__dict__["attack_metric_success_thresh"] = 0.6
-        self.__dict__["attack_metric_success_comp_type"] = "gte"
+        self.__dict__["attack_metric_success_name"] = "P_HIGHER_AUC"
```
simonrnss, line 65: I think failfast should be switched _off_ by default. It's a specific use case for researchers to avoid running lots of tests?

Comments from others? @jim-smith @rpreen @yolaj-nhs ?

aisdc/attacks/worst_case_attack.py
```diff
@@ -492,7 +498,7 @@ def _get_attack_metrics_instances(self) -> dict:
         attack_metrics_experiment = {}
         attack_metrics_instances = {}
 
-        for rep in range(len(self.attack_metrics)):
+        for rep, _ in enumerate(self.attack_metrics):
             attack_metrics_instances["instance_" + str(rep + 1)] = self.attack_metrics[
```
simonrnss: why the +1 here? The json is entirely for machine consumption, and this will mean all developers will have to +- 1?

aisdc/attacks/worst_case_attack.py
```diff
@@ -508,10 +514,10 @@ def _get_dummy_attack_metrics_experiments_instances(self) -> dict:
...
         dummy_attack_metrics_experiments = {}
 
-        for exp_rep in range(len(self.dummy_attack_metrics)):
+        for exp_rep, _ in enumerate(self.dummy_attack_metrics):
             temp_dummy_attack_metrics = self.dummy_attack_metrics[exp_rep]
             dummy_attack_metric_instances = {}
-            for rep in range(len(temp_dummy_attack_metrics)):
+            for rep, _ in enumerate(temp_dummy_attack_metrics):
                 dummy_attack_metric_instances[
                     "instance_" + str(rep + 1)
```
simonrnss, line 522: ditto

aisdc/attacks/worst_case_attack.py
```diff
@@ -540,13 +546,20 @@ def make_report(self) -> dict:
             "dummy_attack_experiments_logger"
         ] = self._get_dummy_attack_metrics_experiments_instances()
 
+        output_pdf = {}
```
simonrnss, line 549: confusing filename for a dictionary? how about `output_for_pdf`?

aisdc/attacks/worst_case_attack.py
```diff
@@ -540,13 +546,20 @@ def make_report(self) -> dict:
...
+        output_pdf["metadata"] = self.metadata
+
         if self.args.report_name is not None:
             json_report = report.create_json_report(output)
             with open(f"{self.args.report_name}.json", "w", encoding="utf-8") as f:
                 f.write(json_report)
 
-        #    pdf = report.create_mia_report(output)
-        #    pdf.output(f"{self.args.report_name}.pdf", "F")
+            pdf = report.create_mia_report(output_pdf)
```
simonrnss, line 561: probably a legacy from me, but `pdf` is also a bad filename :-). `pdf_output`?

aisdc/attacks/worst_case_attack.py
```diff
@@ -791,47 +796,51 @@ def main():
...
         dest="attack_metric_success_comp_type",
         help=(
-            "for computing attack success/failure based on metric_fail_thresh option. Default = %(default)s"
+            """for computing attack success/failure based on
+            --attack-metric-success-thresh option. Default = %(default)s"""
         ),
     )
 
     attack_parser.add_argument(
         "--attack-metric-success-count-thresh",
```
simonrnss, line 835: in the class it's counting failures, so should this param be fail_count or similat?

aisdc/attacks/failfast.py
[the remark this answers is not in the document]
yolaj-nhs: Simon - you mentioned typing in the review comments, do you mean this function signature should become:

`def check_attack_success(self, metric_dict: dict) -> [return_type]:`

?

aisdc/attacks/worst_case_attack.py
yolaj-nhs: Just thinking through the logic of this...

We're implementing fail fast to act as a shortcut to a quick answer. If a user doesn't know the code very well, do we want it to run quickly or run to completion?

I'd argue we want the code to run to completion, and if they know the code well enough and know what they're looking for, they can turn on fail-fast to save themselves time

If failfast defaults to TRUE, in theory a user could run this code for their entire project and never know there's bits of information they're not seeing

That's just my thinking though - feel free to disagree!

aisdc/attacks/worst_case_attack.py
yolaj-nhs: @simonrnss I'm reviewing your review now apparently! I disagree with `pdf_output` as a name - since the dictionary is called `output_pdf` - those names are too similar and could definitely get confusing!

`pdf_report` instead?

aisdc/attacks/failfast.py
```diff
@@ -0,0 +1,96 @@
...
+        Notes
+        -----
+        If value of a given metric value has a value meeting the threshold based on
+        the comparison type returns true otherwise it returns false. This function
+        also counts how many times the attack was successful (i.e. true) and
+        how many times it was not successful (i.e. false).
+        """
+        metric_value = metric_dict[self.metric_name]
+        success_status = False
+        if self.comp_type == "lt":
```
yolaj-nhs, line 39: This is a lot of repeated code - something like this could be cleaner:

`if comp_type == 'lt':
    comparison_function = lambda x, y: x < y
elif comp_type == 'lte':
    comparison_function = lambda x, y: x <= y
elif comp_type == 'gt':
    comparison_function = lambda x, y: x > y
elif comp_type == 'gte':
    comparison_function = lambda x, y: x >= y
elif comp_type == 'eq':
    comparison_function = lambda x, y: x == y
elif comp_type == 'not_eq':
    comparison_function = lambda x, y: x != y

if comparison_function(metric_value, metric_success_thresh) == True:
    success_status = True
    success_count += 1
else:
    success_count = False
    fail_count += 1`

aisdc/attacks/failfast.py
```diff
@@ -0,0 +1,96 @@
...
+    def get_success_count(self):
+        """Returns a count of attack being successful"""
+        return self.success_count
+
+    def get_fail_count(self):
+        """Returns a count of attack being not successful"""
+        return self.fail_count
+
+    def get_attack_summary(self) -> dict:
+        """Returns a dictionary of counts of attack being successful and no successful"""
```
yolaj-nhs, line 92: Typo - should be 'not successful'

shahzadmumtaz22:
Thanks @simonrnss and @yolaj-nhs for a feedback. 

I have made changes in view of your suggestions and left comments in reply. 

After addressing comments I realised that there was merge conflicts (pre-commit test failed). Once I updated merge conflicts (in worst_case_attack.py and worst_case_attack_example.py), I realised worst-case example started to fail. 

On investigation I realised in my PR the way I implemented was like: in a make_report method I was creating two dictionaries (output (dict) - used to dump JSON in new format and out_for_pdf (dict) - a dictionary required to be passed onto report.create_mia_report method with ensuring to meet the previous need of the method)). 

Whereas someone would have made changes to report.create_mia_report method in a merge yesterday (in the development branch) to adopt to a new dictionary (i.e. output, the same used for dumping json)) to be passed on and unpacked within the method of create_mia_report method. Anyways I made changes to only passing the new output dictionary structure (the same used for JSON dumping) to avoid any errors (which means no need for two dictionaries as I was doint it). Though the dictionary passed to create_mia_report has both attack_experiment_logger and dummy_attack_experiment_logger dictionary items but only attack_experiment_logger is unpacked to be used in a pdf report (It sounds like unpacking dummy_attack_experiment_logger is not done and therefore would not be reflected in the pdf report - if that is intentional that is fine otherwise we will have to fix it).

aisdc/attacks/failfast.py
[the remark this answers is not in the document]
simonrnss: yep

simonrnss (approved):
Ok, thanks. All good from me. Will need to just await @yolaj-nhs's approval before merging

aisdc/attacks/failfast.py
```diff
@@ -0,0 +1,92 @@
...
+        """A function to check if attack was successful for a given metric
+
+        Parameters
+        ----------
+        metric_dict: dict
+            a dictionary with all computed metric values
+
+        Returns
+        -------
+        success_status: boolean
```
rpreen, line 34: The return type in the docstring should be `bool` not `boolean`

tests/test_failfast.py
```diff
@@ -0,0 +1,155 @@
...
+    metrics["P_HIGHER_AUC"] = 0.05
+
+    # Option 1
+    args = worst_case_attack.WorstCaseAttackArgs(
+        attack_metric_success_name="P_HIGHER_AUC",
+        attack_metric_success_thresh=0.04,
+        attack_metric_success_comp_type="lte",
+    )
+    failfast_Obj = failfast.FailFast(args)
+    assert failfast_Obj.check_attack_success(metrics) is False
```
rpreen, line 23: `is False` should be `assert not`

tests/test_failfast.py
```diff
@@ -0,0 +1,155 @@
...
+
+    # Option 4
+    args = worst_case_attack.WorstCaseAttackArgs(
+        attack_metric_success_name="P_HIGHER_AUC",
+        attack_metric_success_thresh=0.06,
+        attack_metric_success_comp_type="lt",
+    )
+    failfast_Obj = failfast.FailFast(args)
+    assert failfast_Obj.check_attack_success(metrics) is True
```
rpreen, line 50: unnecessary `is True`

rpreen: and all the way through...

yolaj-nhs (approved):
Approved on my end, pending changes requested from Richard

shahzadmumtaz22:
@rpreen 

Thanks for reviewing. I have made all changes as suggested.

rpreen (approved):

PR #111, opened 2023-04-19
metrics.py refactor

yolaj-nhs:
Changes made:

- Moved metrics.py up a level (from 'attacks' to main aisdc folder)
- Separated the 'get probabilities' and 'get metrics' functionality, to avoid users needing to pass in training data and a model to calculate the metrics. This means the get_metrics function can be used to calculate the metrics for any model which we have prediction labels and true labels
- Added a separate 'get_probabilities' function which takes in a model and dataset and calculates the predicted probabilities (this was originally done inside get_metrics)
- Taken the min_max_disc function out of mia_extremecase.py and removed this file, since that function was the only thing inside the file. Happy to revert this change if people disagree with this!
- Updated all function calls throughout aisdc to match the new metrics.py function syntax


jim-smith:
@yolaj-nhs this is failing the Ci tests.
Having quick look at th code, ther are a few bits that probably coding purists might not like e.g.
if permute_rows == True and len(y_test) == 0:

could just be:
if permute_rows  and len(y_test) == 0:

yolaj-nhs:
Quick note: I'm aware tests are failing, I'm in the process of fixing these

jim-smith:
If you can't force the exceptions to kick in in your tests, please exempt those lines from the code coverage tests so we know we have maintained 100% coverage of the main code.

- it took me tens of hours after GRAIMatter had finished writing tests to reach that target.

simonrnss (approved):
All looks great to me -- happy for this to be merged in

simonrnss:
> All major changes are included as the top comment in this PR. Something important which might need reverted, I've taken the min_max_disc function out of mia_extremecase.py and removed this file, since that function was the only thing inside the file. Happy to revert this change if people disagree with this!

I think this makes sense -- fits better into metrics.py

aisdc/metrics.py
```diff
@@ -155,6 +227,50 @@ def get_metrics(  # pylint: disable=too-many-locals
...
+    -------
+    y_pred_proba: a list of probabilities for each sample in the dataset
+
+    Notes
+    -----
+    If permute_rows is set to true, y_test must also be supplied.
+    The function will then return both the predicted probabilities and corresponding y_test
+    """
+
+    if permute_rows is True and y_test is None:
```
rpreen, line 242: `is True`  :cry:

aisdc/metrics.py
```diff
@@ -155,6 +227,50 @@ def get_metrics(  # pylint: disable=too-many-locals
...
+        N, _ = X_test.shape
+        order = np.random.RandomState(  # pylint: disable = no-member
+            seed=10
+        ).permutation(N)
+        X_test = X_test[order, :]
+        y_test = y_test[order]
+
+    y_pred_proba = clf.predict_proba(X_test)
+
+    if permute_rows is True:
```
rpreen, line 255: `is True` :cry:

rpreen (changes requested):
just a few minor things to clean up...

rpreen (dismissed):
just a few minor things to clean up...

aisdc/metrics.py
```diff
@@ -136,17 +213,12 @@ def auc_p_val(auc: float, n_pos: int, n_neg: int) -> tuple[float, float]:
     return auc_p, auc_std
 
 
-def get_metrics(  # pylint: disable=too-many-locals
-    clf,
-    X_test: np.ndarray,
-    y_test: np.ndarray,
-    permute_rows: bool = True,
+def get_probabilities(  # pylint: disable=too-many-locals
+    clf, X_test: np.ndarray, y_test: np.ndarray = None, permute_rows: bool = None
```
rpreen, line 217: Shouldn't `permute_rows` be default set to `False` and not `None` ? None is not a bool?

rpreen: also `y_test: np.ndarray = None` -- None is not an ndarray

Issue #112, opened 2023-04-25
BUG: Programmatic fails to create report when `n_dummy_reps=0`


simonrnss:
# Introduction

Programmatic call to the worst case attack crashes when attempting to make a report if no dummy repetitions are done.

## Code to reproduce

```python
from sklearn.ensemble import RandomForestClassifier
from sklearn.model_selection import train_test_split

from aisdc.preprocessing.loaders import get_data_sklearn
from aisdc.attacks.dataset import Data
from aisdc.attacks.worst_case_attack import WorstCaseAttack, WorstCaseAttackArgs


# Get a dataset
X, y = get_data_sklearn('nursery')
# Train / test split
trainX, testX, trainy, testy = train_test_split(X, y, test_size=0.95)
# train a model
rf = RandomForestClassifier(min_samples_leaf=2)
rf.fit(trainX, trainy)

# Make data object
nursery = Data()
nursery.add_processed_data(trainX, trainy, testX, testy)

# Run attacks
args = WorstCaseAttackArgs(n_dummy_reps=0, report_name=None)
wca = WorstCaseAttack(args)
wca.attack(nursery, rf)

# Make the output -- crash here!
output = wca.make_report()

```

Code will work fine if `n_dummy_reps` in the constructor for the `args` object is non-zero.

Trace:

```
TypeError                                 Traceback (most recent call last)
[c:\Users\simonr04\git\AI-SDC\scratch\train_models.py](file:///C:/Users/simonr04/git/AI-SDC/scratch/train_models.py) in line 36
     [38](file:///c%3A/Users/simonr04/git/AI-SDC/scratch/train_models.py?line=37) wca.attack(nursery, rf)
     [40](file:///c%3A/Users/simonr04/git/AI-SDC/scratch/train_models.py?line=39) # Make the output -- crash here!
---> [41](file:///c%3A/Users/simonr04/git/AI-SDC/scratch/train_models.py?line=40) output = wca.make_report()

File [c:\Users\simonr04\git\AI-SDC\aisdc\attacks\worst_case_attack.py:432](file:///C:/Users/simonr04/git/AI-SDC/aisdc/attacks/worst_case_attack.py:432), in WorstCaseAttack.make_report(self)
    [430](file:///c%3A/Users/simonr04/git/AI-SDC/aisdc/attacks/worst_case_attack.py?line=429) output["attack_metrics"] = self.attack_metrics
[...]

PR #113, opened 2023-04-25
Fixing dummy_repos bug

yolaj-nhs:
Issue #112 - make_report fails when n_dummy_reps is 0

Solution:
Make dummy_attack_metrics an empty list within attack_from_preds

In _get_global_metrics, check if attack_metrics is an empty array, and return the dictionary without any extra parameters if so


simonrnss (approved):
All good -- thanks

PR #115, opened 2023-04-27
fixing the json file produced by safemodels

mahaalbashir:


mahaalbashir:
@rpreen can you please help me find out why the tests are failing?

jim-smith:
Yes,
I have some memories that we had to go through a few hoops making sure that the outputs from different types of classifiers was serialisable.

simonrnss:
@shahzadmumtaz22 you might need to write something extra into the json encoder, depending on what the object is that is causing the issue

jim-smith:
@rpreen Think this is ready to be approved and merged

PR #116, opened 2023-04-28
Json to recommendation

yolaj-nhs:
Note: this PR is a draft - I still need to add tests and fix linting

Add a generate_report.py file which will take in a JSON file (currently a filename, exact format yet to be decided) and output a formatted text file with recommendations to TREs


yolaj-nhs:
[results.txt](https://github.com/AI-SDC/AI-SDC/files/11352460/results.txt)
[json_output.txt](https://github.com/AI-SDC/AI-SDC/files/11352461/json_output.txt)
Example input and example output attached

aisdc/postprocessing/generate_report.py
```diff
@@ -0,0 +1,283 @@
...
+        raise NotImplementedError()
+
+
+class FinalRecommendationModule(AnalysisModule):
+    """
+    Module that generates the first layer of a recommendation report
+    """
+
+    def __init__(self, report: dict):
+        self.P_VAL_THRESH = 0.1
```
simonrnss, line 30: prob 0.05 is a better default. All arbitrary though.

aisdc/postprocessing/generate_report.py
```diff
@@ -0,0 +1,283 @@
...
+        self.MEAN_AUC_THRESH = 0.65
+
+        self.SVM_WEIGHTING_SCORE = 5
```
simonrnss, line 33: need some thought of how to encode these in a general way as they could easily become cumbersome -- no need to change now though.

aisdc/postprocessing/generate_report.py
```diff
@@ -0,0 +1,283 @@
...
+        self.MIN_SAMPLES_LEAF_SCORE = 3
+        self.STATISTICALLY_SIGNIFICANT_SCORE = 2
+        self.MEAN_AUC_SCORE = 4
+
+        self.report = report
+
+        self.scores = []
+        self.reasons = []
+
+    def _is_svm(self, svm_weighing_score):
```
simonrnss, line 43: So, something to check (not now). Jim et al made a SAFE SVC that wraps a SVC trained with differentially private training. So just need to ensure that the self.report["model"] field isn't SVC in that (i can see how python inheritance might mean that it is).

aisdc/postprocessing/generate_report.py
```diff
@@ -0,0 +1,283 @@
...
+
+    def _rf_min_samples_leaf(self, min_samples_leaf_score):
+        if "model_params" in self.report:
+            if "min_samples_leaf" in self.report["model_params"]:
+                min_samples_leaf = self.report["model_params"]["min_samples_leaf"]
+                if min_samples_leaf < 5:
+                    self.scores.append(min_samples_leaf_score)
+                    self.reasons.append("Min samples per leaf < 5")
+
+    def _statistically_significant_auc(
```
simonrnss, line 59: worth just calling the other module and then parsing it's output?
e.g.
make an instance of the statistically significant AUC module here and the use its output rather than re-computing?

aisdc/postprocessing/generate_report.py
yolaj-nhs: Agreed - some kind of dictionary (e.g. another JSON file) with default parameters was my initial thought, not urgent for now though

aisdc/generate_report.py
```diff
@@ -0,0 +1,312 @@
...
+                ][i]["metrics_array"]
+                if instance["P_HIGHER_AUC"] < p_val_thresh:
+                    stat_sig_auc.append(instance["AUC"])
+
+            n_instances = len(
+                self.report["attack_experiment_logger"]["attack_instance_logger"]
+            )
+            if (
+                len(stat_sig_auc) / n_instances > 0.1
+            ):  # > 10% of AUC are statistically significant
```
simonrnss, line 82: Tiny thing here -- no multple test correction happening here, so we'd expect 1 in 20 attacks to give a significant result by chance. If 1 in 20 happen by chance, then exceeding the 0.1 threshold is probably not that unlikely? Don't need to change this now, but it might give confusing output if the module below says nothing is significant after correcting for multiple testing and this says that there's a problem?

simonrnss (approved):
I'm happy for these to be merged in, assuming that we'll come back and write tests ASAP after we get the release and send stuff to TRE

aisdc/generate_report.py
jim-smith: @yolaj-nhs @simonrnss @rpreen 
Something odd has suddenly started happening with what is obviously a new version of tensorflow_privacy.
I will update safekearas which should fix the lining and test issues

PR #118, opened 2023-04-28
add LIRA fail-fast mechanism and load attack parameters from a config file

shahzadmumtaz22:
1. In view of our discussion, this PR focuses on a fail-fast option not in the same way as implemented in the worst-case attack rather it controls a number of shadow models and degree of freedom based on the confidences computed for the shadow data rows.   
### Programmatically 
The following code explains (a hybrid approach) to pass parameters, the configuration file's name, and a couple of new parameters (shadow_models_fail_fast and n_shadow_rows_confidences_min). These parameters control the number of shadow models trained for LIRA attack scenarios (in the case of setting a value of TRUE for shadow_models_fail_fast parameter, it enables failing fast using the value passed for n_shadow_rows_confidences_min or the default value (which is 2) if not passed this parameter.  
image

### Command line
An example of a fail-fast option using command line,
image

2. The existing JSON output (in the case of LIRA) presents JSON filename only  (like config.json) in the metadata section, 
image

So I changed it to demonstrate all the values in the config.json file as part of the JSON output in the metadata section (see the following figure).
image
   
3. This PR generates JSON output in the new structure suggested by Simon and Yola (see below).
image
At the end of the JSON output, it also demonstrates a number of shadow models trained (see below; in this case it will be 100 as set initially (and demonstrated above in the metadata section) however, if a fail-fast option is enabled this would have been different. 
image


aisdc/attacks/likelihood_attack.py
```diff
@@ -548,6 +622,21 @@ def main():
         default=P_THRESH,
         help=("Significance threshold for p-value comparisons. Default = %(default)f"),
     )
+
+    parser.add_argument(
+        "--shadow-models-fail-fast",
+        action="store",
+        type=parse_boolean_argument,
+        default=True,
```
rpreen, line 630: Instead of having a command line argument passed as `--shadow-models-fail-fast True` and `--shadow-models-fail-fast False` why not just have the default behaviour as `False` and then set the argument to `True` if `--shadow-models-fail-fast` is present?

Just need to change this to something like:
```python
    parser.add_argument(
        "--shadow-models-fail-fast",
        action="store_true",
        required=False,
        dest="shadow_models_fail_fast",
        help=(
            """To stop training shadow models early based on minimum number of
            confidences across all rows (--n-shadow-rows-confidences-min)
            in the shadow data."""
        ),
    )
```

rpreen (changes requested):
Just some minor comments...

tests/test_lira_attack.py
```diff
@@ -28,6 +28,12 @@ def clean_up(name):
         os.remove(name)
 
 
+def test_parse_boolean_argument():
+    """tests parse boolean argument function"""
+    assert likelihood_attack.parse_boolean_argument("true") is True
+    assert likelihood_attack.parse_boolean_argument("false") is False
```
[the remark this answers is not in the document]
shahzadmumtaz22: Updated as suggested.

examples/lira_attack_example.py
```diff
@@ -129,7 +135,16 @@
...
 
+# os.system(
+#     "python -m aisdc.attacks.likelihood_attack run-attack "
+#     "--json-file config.json "
+#     "--report-name example_lira_report "
+#     "--n-shadow-models 100 "
+#     "--shadow-models-fail-fast True "
+#     "--n-shadow-rows-confidences-min 5 "
+# )
+
```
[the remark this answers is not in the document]
shahzadmumtaz22: Removed commented-out code but kept only a couple of lines to demonstrate how to use fail-fast are kept.

aisdc/attacks/likelihood_attack.py
```diff
@@ -92,12 +109,20 @@ def get_args(self) -> dict:
         """Return arguments"""
         return self.__dict__
 
+    def construct_dictionary_from_config_json_file(self, config_filename) -> None:
```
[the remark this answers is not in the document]
shahzadmumtaz22: Type specified for config_filename

aisdc/attacks/likelihood_attack.py
shahzadmumtaz22: The default value is False when an argument is not passed.

rpreen (approved):

aisdc/attacks/likelihood_attack.py
```diff
@@ -32,6 +35,15 @@
 P_THRESH = 0.05  # default significance threshold
 
 
+def parse_boolean_argument(value: str) -> bool:
```
simonrnss, line 38: in light of @rpreen comments -- is this method needed?

aisdc/attacks/likelihood_attack.py
```diff
@@ -248,8 +273,7 @@ def run_scenario_from_preds(  # pylint: disable = too-many-statements, too-many-
         # Train N_SHADOW_MODELS shadow models
         logger.info("Training shadow models")
         for model_idx in range(self.args.n_shadow_models):
-            if model_idx % 10 == 0:
-                logger.info("Trained %d models", model_idx)
+            logger.info("Trained %d models", model_idx + 1)
```
simonrnss, line 276: this is going to lob out an info line for every model. I think that's too much output. Original was every ten?

aisdc/attacks/likelihood_attack.py
```diff
@@ -296,6 +320,19 @@ def run_scenario_from_preds(  # pylint: disable = too-many-statements, too-many-
                         # catch-all
                         shadow_row_to_confidence[i].append(_logit(0))
 
+            lengths_shadow_row_to_confidence = {
```
simonrnss, line 323: this is the bulk of the failfast code and it's hard to follow without comments -- could you add some? Also missing some overall documentation (in the codebase) about how this failfast works (I think I missed the meeting).

I _think_ what's happening is that you stop when you have at least a mininum number of confidence values for each example (which may save some time due to the random nature of the dataset splitting) but I'm not sure -- needs clarification so that a user will know when to use this.

aisdc/attacks/likelihood_attack.py
```diff
@@ -529,6 +589,20 @@ def main():
         dest="n_shadow_models",
         help=("The number of shadow models to train (default = %(default)d)"),
     )
+
+    parser.add_argument(
+        "--n-shadow-rows-confidences-min",
+        type=int,
+        action="store",
+        dest="n_shadow_rows_confidences_min",
+        default=2,
```
simonrnss, line 598: if my interpretation (above) is right, this is way too low for a sensible default value?

examples/lira_attack_example.py
```diff
@@ -129,7 +135,9 @@
     "python -m aisdc.attacks.likelihood_attack run-attack "
     "--json-file config.json "
     "--report-name example_lira_report "
-    "--n-shadow-models 100"
+    "--n-shadow-models 100 "
+    # "--shadow-models-fail-fast True "
+    # "--n-shadow-rows-confidences-min 5 "
```
simonrnss, line 140: still too low probably

examples/lira_attack_example.py
```diff
@@ -129,7 +135,9 @@
     "python -m aisdc.attacks.likelihood_attack run-attack "
     "--json-file config.json "
     "--report-name example_lira_report "
-    "--n-shadow-models 100"
+    "--n-shadow-models 100 "
+    # "--shadow-models-fail-fast True "
```
simonrnss, line 139: why not have two examples in the script? One with, and one without these options?

tests/test_lira_attack.py
```diff
@@ -28,6 +28,12 @@ def clean_up(name):
         os.remove(name)
 
 
+def test_parse_boolean_argument():
```
simonrnss, line 31: should we have a test to check the actual behaviour of the new options?

simonrnss (changes requested):
Just some minor things...all looks great in general

aisdc/attacks/likelihood_attack.py
```diff
@@ -32,6 +35,15 @@
 P_THRESH = 0.05  # default significance threshold
 
 
+def parse_boolean_argument(value):
+    """Parses string value to a boolean"""
+    value = value.lower()
+    return_value = False
+    if value in ["true", "True"]:
+        return_value = True
+    return return_value
```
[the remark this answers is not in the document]
rpreen: There is a good reason that there isn't a bool type specified. The absence of a command line option means that it is not active, i.e., it's false. So passing in `--shadow-models-fail-fast` would mean the user wants it to be true. Adding "true" or "false" after that option is redundant. That's why I suggested the add argument code below, and that means we can drop this function because it's bloat.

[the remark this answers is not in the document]
shahzadmumtaz22: Thanks for your feedback. 

Yes, I agree; I did not get it at first. 

Now the function is removed.

aisdc/attacks/likelihood_attack.py
shahzadmumtaz22: Yes, removed the method given feedback from @rpreen

aisdc/attacks/likelihood_attack.py
shahzadmumtaz22: Yes your interpretation is correct, the default is now set to be 5 instead of 2.

aisdc/attacks/likelihood_attack.py
shahzadmumtaz22: Added some comments to explain the process.

examples/lira_attack_example.py
shahzadmumtaz22: In the script just added two examples as suggested.

examples/lira_attack_example.py
shahzadmumtaz22: @simonrnss 

Should we say 10 to be more sensible as a default value?

tests/test_lira_attack.py
shahzadmumtaz22: Added a test to do code coverage. Will add it later today to test the functionality of fail fast.

rpreen (changes requested):

aisdc/attacks/likelihood_attack.py
```diff
@@ -92,12 +102,20 @@ def get_args(self) -> dict:
         """Return arguments"""
         return self.__dict__
 
+    def construct_dictionary_from_config_json_file(self, config_filename: str) -> None:
+        """Return a dictionary object reading through a config.json"""
```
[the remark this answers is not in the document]
rpreen: I'm starting to wonder why we even need these special attack args classes for each attack and whether this should be just handled within the attack classes?

rpreen (approved):

simonrnss (approved):

PR #119, opened 2023-04-30
Development shahzad mia configfile

shahzadmumtaz22:
This PR focuses on using a configuration file for programmatic and command-line use of worst-case. The old functionality of using worst-case scenarios without a configuration file is unaffected.  

Need to discuss the LIRA scenario as that has a hybrid approach (passing parameters and using a configuration file in combination). Should we implement that in the same way as I did for the worst-case scenario in this PR?

### DIFF
--- a/aisdc/attacks/worst_case_attack.py
+++ b/aisdc/attacks/worst_case_attack.py
@@ -427,9 +427,9 @@ class WorstCaseAttack(Attack):
 
         # Global metrics
         self.metadata["global_metrics"] = self._get_global_metrics(self.attack_metrics)
-        # self.metadata["baseline_global_metrics"] = self._get_global_metrics(
-        # self.dummy_attack_metrics
-        # )
+        self.metadata["baseline_global_metrics"] = self._get_global_metrics(
+        self.dummy_attack_metrics
+        )
 
     def _get_attack_metrics_instances(self) -> dict:
         """Constructs the metadata object, after attacks"""
--- a/aisdc/attacks/worst_case_attack.py
+++ b/aisdc/attacks/worst_case_attack.py
@@ -81,7 +81,7 @@ class WorstCaseAttackArgs:
         with open(config_filename, encoding="utf-8") as f:
             config = json.loads(f.read())
         for _, k in enumerate(config):
-            self.__dict__[k] = config[k]            
+            self.__dict__[k] = config[k]
 
 
 class WorstCaseAttack(Attack):
--- a/aisdc/attacks/worst_case_attack.py
+++ b/aisdc/attacks/worst_case_attack.py
@@ -54,7 +54,7 @@ class WorstCaseAttackArgs:
             "min_samples_leaf": 10,
             "max_depth": 5,
         }
-        self.__dict__["json_file"] = None 
+        self.__dict__["json_file"] = None
         self.__dict__.update(kwargs)
         # Reading parameters from a json file
         if self.__dict__["json_file"] is not None:
@@ -81,8 +81,8 @@ class WorstCaseAttackArgs:
         with open(config_filename, encoding="utf-8") as f:
             config = json.loads(f.read())
         for _, k in enumerate(config):
-            self.__dict__[k] = config[k]
-            
+            self.__dict__[k] = config[k]            
+
 
 class WorstCaseAttack(Attack):
     """Class to wrap the worst case attack code"""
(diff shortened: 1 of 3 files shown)


aisdc/attacks/worst_case_attack.py
```diff
@@ -52,7 +54,17 @@ def __init__(self, **kwargs):
             "min_samples_leaf": 10,
             "max_depth": 5,
         }
+        self.__dict__["json_file"] = None
         self.__dict__.update(kwargs)
+        # Reading parameters from a json file
+        if self.__dict__["json_file"] is not None:
+            print("First If passed")
```
rpreen, line 61: If this print statement is needed it would be best put in a logger?

shahzadmumtaz22: Supposed to remove it as I put it to do some test and removed it now.

tests/test_worst_case_attack.py
```diff
@@ -20,6 +21,57 @@ def clean_up(name):
         os.remove(name)
 
 
+def test_config_file_arguments_parsin():
```
rpreen, line 24: These tests don't have any assertions so aren't actually checking anything?

Also, the OS calls to execute the command line wont work on linux (including the CI runner) because it needs to call `python3` and not `python` - these really should have been updated in all the examples too but that's an issue beyond this PR.

aisdc/attacks/worst_case_attack.py
```diff
@@ -67,7 +69,16 @@ def __init__(self, **kwargs):
         self.__dict__["attack_metric_success_comp_type"] = "lte"
         self.__dict__["attack_metric_success_count_thresh"] = 5
         self.__dict__["attack_fail_fast"] = False
+        self.__dict__["json_file"] = None
```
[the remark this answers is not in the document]
simonrnss: I've just labeled it here, but think it should change throughout

examples/worst_case_attack_example.py
```diff
@@ -50,6 +51,23 @@
 test_preds = target_model.predict_proba(test_X)
 
 # [TRE] Define some attack parameters
+# Use of configuration file to load parameters
```
simonrnss, line 54: Why comment it all out? Better to have multiple examples within the script that all run I think? These scripts also work as regression tests so best that they run.

tests/test_worst_case_attack.py
simonrnss: @rpreen can you make an issue to address this please?

tests/test_worst_case_attack.py
```diff
@@ -20,6 +21,57 @@ def clean_up(name):
...
+        "p_thresh": 0.05,
+        "test_prop": 0.5,
+        "report_name": "programmatically_worstcase_report",
+    }
+    with open("config_worstcase_test.json", "w", encoding="utf-8") as f:
+        f.write(json.dumps(config))
+    _ = worst_case_attack.WorstCaseAttackArgs(
+        json_file="config_worstcase_test.json",
+    )
+    os.remove("config_worstcase_test.json")
```
simonrnss, line 38: add assertions here that the attributes within the args object within the attack are actually set at the values in the config dict (and therefore ensure that these are different from the defaults to test that the changes are actually picked up)

aisdc/attacks/worst_case_attack.py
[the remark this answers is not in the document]
shahzadmumtaz22: Made changes for LIRA attack as well (in PR 118)

examples/worst_case_attack_example.py
shahzadmumtaz22: Updated script to run all examples run.

examples/worst_case_attack_example.py
```diff
@@ -164,6 +182,26 @@
 os.system("python -m aisdc.attacks.worst_case_attack run-attack --help")
 
 # [TRE] Then they run the attack
+# Worstcase attack by passing a configuratation file for loading parameters
```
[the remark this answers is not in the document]
shahzadmumtaz22: Ensured all examples run.

tests/test_worst_case_attack.py
shahzadmumtaz22: Assertions are added as suggested.

rpreen (approved):

PR #120, opened 2023-05-01
standardise MIA attack output

shahzadmumtaz22:
Rebased this branch from the development and fingers crossed should not have any merge conflicts. 

The PR focuses on standardising the variable names across worst_case_attack and LIRA attacks. This was described in an issue https://github.com/AI-SDC/AI-SDC/issues/109 and brief detail of this standarisation exercise is given in the shared excel file:(https://dmail-my.sharepoint.com/:x:/g/personal/smumtaz001_dundee_ac_uk/ERkQkNOtYdRKtsOwRvA-dLwBUFsAqq5DO99UpusP5FvL2g?e=FWj1nO). The excel file has three sheets: MIA_metadata_standardisation, MIA_MetricOutputStandardisation and MIA_Process_standardisation.


rpreen:
In future, we really should keep each PR narrowly scoped to a single feature/issue and not have crazy dependencies on other PRs that are merged into each other because it makes the reviewing process much more difficult and they get out of synch. I seem to be seeing things here that were present in the other PRs that were subsequently removed from them.

aisdc/attacks/worst_case_attack.py
```diff
@@ -44,8 +44,8 @@ def __init__(self, **kwargs):
         self.__dict__["test_prop"] = 0.3
         self.__dict__["n_rows_in"] = 1000
         self.__dict__["n_rows_out"] = 1000
-        self.__dict__["in_sample_filename"] = None
-        self.__dict__["out_sample_filename"] = None
+        self.__dict__["training_preds_file"] = None
```
simonrnss, line 47: please change the key to make it clear this is a file _name_

simonrnss (changes requested):
I'm not sure what the rationale for these changes are, sorry. I don't think we should standardise output to the extent that we mash things together that don't fit together. A likelihood attack is quite a different thing to a worst case attack, and they will out of necessity produce different outputs.

Perhaps some clearer rationale of what the purpose of the standardisation here is?

shahzadmumtaz22:
> I'm not sure what the rationale for these changes are, sorry. I don't think we should standardise output to the extent that we mash things together that don't fit together. A likelihood attack is quite a different thing to a worst case attack, and they will out of necessity produce different outputs.
> 
> Perhaps some clearer rationale of what the purpose of the standardisation here is?

Thanks @simonrnss for the feedback. The rationale was to use standardised variable names across both types of attacks where possible. For example, in the worst_case_attack we use in_sample_filename and out_sample_filename and in the lira_attack for the same purpose we use variable names as training_preds_file and test_preds_file respectively. I think we should use the same name across like training_preds_filename and test_preds_filename.
image

This name standardisation will not only be helpful to have the same name appearing in outputs generated from both types of attacks making it easier to interpret but also using the same names when writing scripts to run attacks.

aisdc/attacks/worst_case_attack.py
shahzadmumtaz22: Updated the names to "training_preds_filename" and "test_preds_filename".

aisdc/attacks/likelihood_attack.py
```diff
@@ -498,8 +498,8 @@ def setup_example_data(self) -> None:
             "testing_data_file": "test_data.csv",
             "training_preds_file": "train_preds.csv",
             "testing_preds_file": "test_preds.csv",
-            "target_model": ["sklearn.ensemble", "RandomForestClassifier"],
-            "target_hyppars": {"min_samples_split": 2, "min_samples_leaf": 1},
+            "mia_attack_model": ["sklearn.ensemble", "RandomForestClassifier"],
+            "mia_attack_model_hyp": {"min_samples_split": 2, "min_samples_leaf": 1},
```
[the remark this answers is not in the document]
shahzadmumtaz22: Thanks @simonrnss. I agree in LIRA there is no attack model rather there is a shadow model to be trained which should have the same specification as the target model. So left the names as they were originally present but made a minor change by changing the variable name target_hyppars to target_model_hyp.

simonrnss (approved):

simonrnss:
Ok, thanks -- happy now.

rpreen (approved):

Issue #117, opened 2023-04-28
generate_reports.py needs tests


yolaj-nhs:
generate_reports.py (PR #116) needs unit tests

Code has been tested, but unit tests are needed for completeness

simonrnss:
@yolaj-nhs Is this done?

yolaj-nhs:
> @yolaj-nhs Is this done?

Yes - closed, thanks for the reminder!

PR #122, opened 2023-05-02
Adding tests to generate_report.py

yolaj-nhs:
Relates to issue #117


tests/test_generate_report.py
```diff
@@ -0,0 +1,183 @@
...
+        _ = f.process_dict()
+
+    def test_complete_runthrough(self):
+        """test the full process_json file end-to-end when valid parameters are passed"""
+        json_formatted = self.get_test_report()
+        _ = self.process_json_from_file(json_formatted)
+
+    def test_cleanup(self):
+        """gets rid of files created during tests"""
+        names = ["test.json", "results.txt", "1024-WorstCase attack.png"]
```
simonrnss, line 181: this might be my fault, but could we fix the module that plots to not produce filenames with whitespaces?

tests/test_generate_report.py
```diff
@@ -0,0 +1,183 @@
...
+        """test the SummariseUnivariateMetricsModule"""
+        json_formatted = self.get_test_report()
+        f = SummariseUnivariateMetricsModule(json_formatted)
+        _ = f.process_dict()
+
+    def test_auc_pvals_module(self):
+        """test the SummariseAUCPvalsModule"""
+        json_formatted = self.get_test_report()
+        f = SummariseAUCPvalsModule(json_formatted)
+        _ = f.process_dict()
```
simonrnss, line 149: Happy with this for now. After this release I think we should actually test the output of these. If we have a static json output produced by #126 within the repo, we can use it to test that the modules are also giving the correct answers (and it will also pick up any module output formatting changes that creep in).

simonrnss (approved):
All good, thanks

tests/test_generate_report.py
yolaj-nhs: Created issue #130 to cover this

rpreen (approved):

PR #123, opened 2023-05-02
Turning off non-default report logs

yolaj-nhs:
The generate_report.py script contains different modules for reporting different report metrics. Make the default behaviour include only the simplest one

### DIFF
--- a/aisdc/generate_report.py
+++ b/aisdc/generate_report.py
@@ -241,11 +241,11 @@ class LogLogROCModule(AnalysisModule):
         ].values()
         all_tpr = np.zeros((len(metrics), len(base_fpr)), float)
         for i, metric_set in enumerate(metrics):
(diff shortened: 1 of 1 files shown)


aisdc/generate_report.py
```diff
@@ -291,22 +291,10 @@ def process_json(input_filename: str, output_filename: str):
 
     modules = [
         FinalRecommendationModule(json_report),
-        SummariseUnivariateMetricsModule(json_report),
-        SummariseAUCPvalsModule(json_report, p_thresh=0.05),
```
simonrnss, line 295: Maybe we can quickly discuss -- I think there is a risk that in this first example, if we simplify things too much we paint a picture for TREs that things are easy...will discuss. Appreciate that we need to get it finalised now.

simonrnss: Actually, ignore this. No time at the moment. However, I'll make an issue that we should refactor a bit so that your module re-uses code from some of the others.

simonrnss (approved):

PR #124, opened 2023-05-02
Bugfix sr example scripts

simonrnss:
@shahzadmumtaz22 pointed out that the merge on Friday broke the worst case attack example. This PR fixes that break so should be merged prior to the release.

I have tested all four examples, and the other three work fine (Windows).

### DIFF
--- a/aisdc/attacks/worst_case_attack.py
+++ b/aisdc/attacks/worst_case_attack.py
@@ -427,9 +427,9 @@ class WorstCaseAttack(Attack):
 
         # Global metrics
         self.metadata["global_metrics"] = self._get_global_metrics(self.attack_metrics)
-        # self.metadata["baseline_global_metrics"] = self._get_global_metrics(
-        # self.dummy_attack_metrics
-        # )
+        self.metadata["baseline_global_metrics"] = self._get_global_metrics(
+            self.dummy_attack_metrics
+        )
 
     def _get_attack_metrics_instances(self) -> dict:
         """Constructs the metadata object, after attacks"""
--- a/examples/worst_case_attack_example.py
+++ b/examples/worst_case_attack_example.py
@@ -21,10 +21,7 @@ python -m examples.worst_case_attack_example
 
 """
 import os
-import os
-import sys
 
-sys.path.append(os.path.dirname(os.path.dirname('__file__')))
 import numpy as np
 from sklearn.datasets import load_breast_cancer
 from sklearn.model_selection import train_test_split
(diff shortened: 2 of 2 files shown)


yolaj-nhs (approved):

rpreen:
Worst case example doesn't seem to be producing a PDF - is that to be expected? The others do.

simonrnss:
@rpreen yes -- we could make it produce one by setting `report_name` on line 65. The command line version is producing one. In fact, it seems to be setting the param twice: "example_report_risky.pdf" and "example_report.pdf" -- which would make most sense to keep?

rpreen:
AIA example produces: `aia_example.pdf`
LIRA example produces: `example_lira_report.pdf` and `lira_example_report.pdf` 

So having worst case produce `worstcase_example.pdf` or `worstcase_example_report.pdf` might make sense?

rpreen:
I think the PDFs are useful for first-time users because it's something that's easy to open and view - whereas the JSON is bit more cryptic for a first glance.

simonrnss:
ah, ok, pdf report is not currently compatible with the new output format. Just looking how easy it would be to fix

simonrnss:
Ok, fixed. Will make an issue to fix this properly

shahzadmumtaz22:
> ah, ok, pdf report is not currently compatible with the new output format. Just looking how easy it would be to fix
@simonrnss and @rpreen 

Apologies for Jumping in the coversation. 

But in my PRs this is ensured that existing pdf generating functionality should not be affected even when generating new json format outputs both for worst-case and LIRA scenarios - if that helps. I briefly explain the same to Yola on Friday (28 April) in a one to one meeting.

simonrnss:
Hi @shahzadmumtaz22 -- sure, ok, that'll hopefully override these changes once they're merged in. Be v helpful if in future all your prs could be linked to a single issue so that it's a bit more visible what's where. We'll get them all merged once the current rush is over!

simonrnss:
> In the long run, maybe the PDF reporting should be at a higher level: attack produces JSON, analysis module then reads the JSON and produces the summary which could include outputting a PDF? Not something to worry about right now I guess.

I made an issue to fix the .pdf, but you're totally right -- I'll update it to cover this. Basically, a single module to go from json -> pdf.

rpreen:
> Hi @shahzadmumtaz22 -- sure, ok, that'll hopefully override these changes once they're merged in. Be v helpful if in future all your prs could be linked to a single issue so that it's a bit more visible what's where. We'll get them all merged once the current rush is over!

I think we should also make it a policy that all future PRs need to update the `CHANGELOG.md` because going backwards now to fill that in is a little more tricky...

rpreen:
Not sure if I'm doing something wrong but this still doesn't produce a PDF for worst case

PR #151, opened 2023-05-11
149 refactor lira attack test to use fixtures and avoid repetition

simonrnss:
Tests are sometimes not structured optimally, and have quite a lot of code repetition. This is a bit untidy but can also slow things down as the same code is run over and over again when it could just be run once.

In this example I've factored the tests into a class and used the cls methods: `setUpClass` and `tearDownClass` to cover code that can just be run once and cleanup (respectively).

If this example looks good to people, we can start identifying other places where the structure could be improved.


yolaj-nhs (approved):
Looks good to me, I'm happy to make these changes in other test files where relevant

jim-smith:
@simonrnss @rpreen Adding Richard to list of approvers as he will have more input than me on implications of mixing  the  the 'unit test' way- which creates classes, or the pytest way, which I sort of did in a few  places (e.g. test_attacks_via_safemodel.py)

where I had a function that individual tests called, but it 'knew' whether had run already.   Not sure how efficient this was in terms of loading things into memory etc.

simonrnss:
@rpreen have you any comments on this re factoring tests as classes versus methods?

rpreen:
I have a preference for pytest rather than unittest because it has a lot of nice features and also in part because most of the major Python projects like sklearn use it, but it's not a problem if some of them use unittest classes for now (you can also use classes in pytest so there's nothing wrong with that). Obviously the most important bit is just getting good tests in place and pytest will pickup and run both.

It looks like this PR needs to resolve the conflict with the changes that were made in the PR that changed LIRAs outputs and also changed/added tests.

simonrnss:
Thanks @rpreen -- so, happy for this to be merged (once the conflict is resolved)?

rpreen:
yes

yolaj-nhs:
I checked over this a second time after the merge conflict has been fixed - looks good to me

PR #165, opened 2023-05-25
LIRA attack from configuration file and Avoiding Code repetition across LIRA and WorstCase While Loading Configuration File

shahzadmumtaz22:
The PR addresses issue https://github.com/AI-SDC/AI-SDC/issues/161.

1. The LIRA attack from the command line supports the functionality (hybrid parameter approach - by passing JSON config file as an argument and other command line arguments) by initialising all the parameters from the object initiation of LIRAAttackArgs within the _run_attack function (not from the attack_from_config function rather this function reads prediction files, data files and creation of shadow model given parameters in the JSON file). An example of a hybrid parameter passing option is given as: 
image
Why hybrid approach is necessary because target model (shadow model) specification (hyperparameters are not always fixed and we can not pass through a Pythonic dictionary object on the command line and passing through a configuration file can handle this situation). 

2.  This PR also adds a function of using a JSON configuration file only as a parameter to run the LIRA attack (though the above hybrid approach also can handle such a situation if only a configuration JSON file argument is given and it will set all the required parameters in place, but this separate command is added ). An example is demonstrated is given as:
image
However, it is important to mention that the following will work the same way but added a command "run-attack-from-config-file" to be consistent across WorstCase and LIRA attacks. 
image

3.  The liniting has also come up with an error, and @rpreen also pointed out the same thing to avoid repeating the same code across different attacks. So created a class called ConfigFile within attack.py file for loading dictionary items from the given JSON files. This was needed both for WorstCaseAttackArgs and LIRAAttackArgs.


aisdc/attacks/attack.py
```diff
@@ -14,3 +17,20 @@ def attack(self, dataset: Data, target_model: sklearn.base.BaseEstimator) -> Non
...
+
+    def load_config_file_into_dict(self, attack_args_dict: dict) -> dict:
+        """Reads a configuration file and loads it into a dictionary object"""
+        if self.config_filename is not None:
+            if os.path.isfile(self.config_filename):
+                with open(self.config_filename, encoding="utf-8") as f:
+                    config = json.loads(f.read())
+                for _, k in enumerate(config):
+                    attack_args_dict[k] = config[k]
+        return attack_args_dict
```
simonrnss, line 36: if `self.config_filename` is None then this will crash I think? 
Propose removing the if. Object cannot be created without a filename being passed into the constructor.

aisdc/attacks/likelihood_attack.py
```diff
@@ -80,14 +79,20 @@ def __init__(self, **kwargs):
...
+        self.__dict__["attack_config_json_file_name"] = None
         self.__dict__["shadow_models_fail_fast"] = False
-
-        if os.path.isfile(self.__dict__["attack_config_json_file_name"]):
-            self.load_config_file_into_dict(
-                self.__dict__["attack_config_json_file_name"]
-            )
         self.__dict__.update(kwargs)
+        configfile_obj = ConfigFile(self.__dict__["attack_config_json_file_name"])
+        dict_obj = configfile_obj.load_config_file_into_dict(self.__dict__)
```
simonrnss, line 92: won't `self.__dict__` be passed by reference, so the things you add don't then need to be added again to `self.__dict__` with the update statement on line 93?

tests/lrconfig2.json
simonrnss, line 1: please use a clearer filename

simonrnss (changes requested):
Great, thanks. Some minor comments.

aisdc/attacks/attack.py
shahzadmumtaz22: Thanks, changed as suggested.

aisdc/attacks/likelihood_attack.py
shahzadmumtaz22: Thanks, I agree and changes made as suggested.

tests/lrconfig2.json
shahzadmumtaz22: Renamed file as lrconfig_cmd to use on the commandline in tests.

aisdc/attacks/attack.py
```diff
@@ -14,3 +17,18 @@ def attack(self, dataset: Data, target_model: sklearn.base.BaseEstimator) -> Non
...
+
+class ConfigFile:  # pylint: disable=too-few-public-methods
+    """ConfigFile class to load parameters from json configuration file"""
+
+    def __init__(self, config_filename):
+        self.config_filename = config_filename
+
+    def load_config_file_into_dict(self, attack_args_dict: dict) -> None:
+        """Reads a configuration file and loads it into a dictionary object"""
+        if os.path.isfile(self.config_filename):
```
simonrnss, line 30: One final change here -- please add something in an `else` for the `if os.path...` to raise a log error. Or, just remove the check and use the fact that python will raise a FileNotFoundException anyway...(I prefer the latter)

simonrnss (approved):
Just one tiny change, but clicking approve as I don't need to see it again after that change!

aisdc/attacks/attack.py
shahzadmumtaz22: Thanks @simonrnss. I have removed the if statement (if os.path...) as you suggested.

shahzadmumtaz22:
> Just one tiny change, but clicking approve as I don't need to see it again after that change!

Thanks @simonrnss. changes made as suggested.

yolaj-nhs (approved):

jim-smith (commented):
do we need a class for ConfigFile, since the only method is to read afield into a dict. ?
Is this not a just a function that could be called if we put into this file (attacks.py) or even somewhere more general purpose

Issue #155, opened 2023-05-16
Streamlining the outputs produced


rpreen:
The following tries to summarise a few existing issues (e.g., #110, #128, #147, #148, #152) to get an overview of what we would like the outputs to be.

As I understand it, the current AI-SDC outputs:

* safemodel `request_release()` produces a JSON checkfile containing a summary description of the model type, results of hyperparameter checks, and attack settings;
* each attack produces an individual JSON file containing its results;
* the `generate_report` module has a function `process_json()` that produces a txt file with a high level summary giving a final recommendation, including a risk score (out of 5) based on a single attack's results (e.g., worst case attack)

Suggestion: `TARGET.json` :heavy_plus_sign:  `ATTACK_CONFIG.json` :arrow_right: `ATTACKS.json` :arrow_right: `SUMMARY.txt`

* `TARGET.json`: a JSON file containing the target model details. For example, paths to the model and data, the name of the model, the hyperparameters, the name of the dataset, the labels, the number of training/test items, number of features and classes, and maybe something about preprocessing --- all the things necessary for running attacks and any extra information that the TRE checker may wish to view *even if it's not part of any automated checks*. Ideally, most/all of this would be automatically extracted, but for other languages such as R this could be hand-crafted or we could produce a helper function to make sure that the bare minimum is included.
* safemodel `request_release()` includes information around model checking in the `TARGET.json`
* `ATTACK.json`: each attack on a single target model (a single `TARGET.json`) should append its parameters and results to a single JSON file; if we also support an `ATTACK_CONFIG.json` then that together with the `TARGET.json` should be enough to execute the attacks and produce the `ATTACK.json`
* `SUMMARY.txt`: a `generate_report` module that processes both `TARGET.json` and `ATTACK.json` to provide a summary including all of the relevant information a TRE checker needs, including the overall recommendation.
* A GUI (WP3) can take all three JSON files and allow the checker to explore the information generated.

simonrnss:
Agree that this is worth doing. Not sure it's a release 1.0.5 thing -- would prefer to keep that for cleaning up fairly vital outstanding things from the previous rushed release? This is quite a sizable piece of work too, which would delay releasing some of the stuff we should have had in the previous release.

jim-smith:
## Master List of Outstanding Tasks:
(if you can't add to this put in a select comment on an owner will transfer them)

rpreen:
Closing this as resolved with v.1.0.6.

PR #171, opened 2023-06-08
rename data class as target and add target model

rpreen:
* Renames `dataset.py` to `target.py`
* Renames `Data` class to `Target` class
* Adds the target model to the `Target` class
* Updates all examples and tests

This is the first step towards generating the `TARGET.json` described in #155


simonrnss (approved):
Looks great -- much cleaner structure, thanks.

PR #173, opened 2023-06-13
Output of each attack is appended to a single JSON file

yolaj-nhs:
Currently, JSON files are created by each attack, which results in multiple JSON files floating around. I've changed all JSON file creation to come from one place, meaning only one file is created, and attack outputs are appended to the end of a single file. Result: one JSON file for all attacks


shahzadmumtaz22:
Hi @yolaj-nhs,

Thanks, I understand this PR focuses on generating a single JSON output file across attacks. 

I struggled to understand this implementation when tested to see a single output file across attacks.

It would be helpful to have an example demonstrating two or more attacks and generating a single JSON attack file. This should be an additional functionality to create a single output file if the user would like, and the feature of generating separate output files across attacks should be supported.

yolaj-nhs:
An example of how this can be used is included in the zip file. Note: the file includes a bunch of setup which can be ignored, and then has the examples at the bottom

[example_json_report_script.zip](https://github.com/AI-SDC/AI-SDC/files/11746624/example_json_report_script.zip)

simonrnss (approved):

rpreen:
Just a minor thought: this requires the researcher to import `GenerateJSONModule` and then instantiate an object with the file name before passing that to the attack. Would it be possible to hide this from the researcher so it's a bit cleaner for them and they don't have to import that class?

So at the moment to write to the same file:
```python
from aisdc.attacks.attack_report_formatter import GenerateJSONModule
...
g = GenerateJSONModule("single_file.json")
...
output = lira_attack_obj.make_report(g)
...
output = attack_obj.make_report(g)
```

Would be a bit cleaner if it was as follows?
```python
filename = "single_file.json"
output = lira_attack_obj.make_report(filename)
...
output = attack_obj.make_report(filename)
```

rpreen:
Wouldn't it be as simple as if the file exists then it just loads it and appends to it? And if that wasn't the desired behaviour the `make_report()` function could have a parameter like `append=False` where it would overwrite.

It's fine though. Just a thought.

PR #174, opened 2023-06-13
Refactor Attack Argument handling Subsuming with Attack as a Dictionary

shahzadmumtaz22:
1. Removed *AttackArgs classes which hold dictionaries for each type of attack
2. Dictionaries of args are now held directly within each  *Attack (i.e. LIRAAttack, WorstCaseAttack and AttributeAtack)
3. Static functions are added to attack.py to load a dictionary for each type of attack initially, with default values assigned to each dictionary item. 
4. Changes made where needed to work with the dictionary available in each type of attack class (For instance, in examples and tests)
5. ConfigFile Class is also removed from attack.py, and a static function is added to load dictionary items from the configuration file
6. Some codecov issues arise, and they are also fixed


rpreen (changes requested):
I think the default argument values should be set in the implementation classes and not in the Attack module. They could be set in the constructors?

I'm also wondering if it might be cleaner if instead of using an args dictionary or kwargs that we just explicitly define the defaults in the constructor and that way it will also be visible in the documentation? Is there any problem with doing that? So for example:

```python
class WorstCaseAttack(Attack):
    """Class to wrap the worst case attack code"""

    def __init__(
        self,
        n_reps: int = 2,
        p_thresh: float = 0.05,
    ) -> None:
        """Constructs an object to execute a worst case attack.

        Parameters
        ----------

        n_reps : int
            Number of attack repetitions to perform.
        p_thresh : float
            Threshold ...
        """
        self.n_reps = n_reps
        self.p_thresh = p_thresh

```

Then later on in the WorstCaseAttack it can be used just like a normal variable:

```python
for rep in range(self.n_reps):
```

instead of:

```python
for rep in range(self.args["n_reps"]):
```

shahzadmumtaz22:
> I think the default argument values should be set in the implementation classes and not in the Attack module. They could be set in the constructors?
> 
> I'm also wondering if it might be cleaner if instead of using an args dictionary or kwargs that we just explicitly define the defaults in the constructor and that way it will also be visible in the documentation? Is there any problem with doing that? So for example:
> 
> ```python
> class WorstCaseAttack(Attack):
>     """Class to wrap the worst case attack code"""
> 
>     def __init__(
>         self,
>         n_reps: int = 2,
>         p_thresh: float = 0.05,
>     ) -> None:
>         """Constructs an object to execute a worst case attack.
> 
>         Parameters
>         ----------
> 
>         n_reps : int
>             Number of attack repetitions to perform.
>         p_thresh : float
>             Threshold ...
>         """
>         self.n_reps = n_reps
>         self.p_thresh = p_thresh
> ```
> 
> Then later on in the WorstCaseAttack it can be used just like a normal variable:
> 
> ```python
> for rep in range(self.n_reps):
> ```
> 
> instead of:
> 
> ```python
> for rep in range(self.args["n_reps"]):
> ```

Thanks @rpreen for the feedback. 

I agree that the default parameter values bet set within each attack class's constructor. So removed static functions from attack.py module, and the parameters are set with default values within the constructor of each attack-specific class.

I have created a dictionary called args to keep all the parameters passed and to update them through kwargs and/or a configuration file. I think putting all the configuration parameters within the args dictionary seems cleaner when creating a report so we can only dump an args dictionary directly within (For example, see a screenshot of the report where we use the args dictionary) 
image

rpreen:
Yeah, I see what you mean.

I'm looking at how sklearn does their `get_params()` and it looks like they inspect the constructor to pull out the parameter names and then extract only those from the dict. Not sure if it's worth the fuss for us to do.

https://github.com/scikit-learn/scikit-learn/blob/21312644df0a6b4c6f3c27a74ac9d26cf49c2304/sklearn/base.py#L144

rpreen:
This looks much cleaner.

If the following functions are included in the attack class, `get_params()` should return the parameters and their values.

```python
import inspect


    @classmethod
    def _get_param_names(cls):
        """Get parameter names"""
        init_signature = inspect.signature(cls.__init__)
        parameters = [
            p.name
            for p in init_signature.parameters.values()
            if p.name != "self" and p.kind != p.VAR_KEYWORD
        ]
        return parameters

    def get_params(self):
        """
        Get parameters for this attack.

        Returns
        -------
        params : dict
            Parameter names mapped to their values.
        """
        out: dict = {}
        for key in self._get_param_names():
            out[key] = getattr(self, key)
        return out
```

So instead of having to define the keys to exclude and calling `_exclude_keys_from_dict()` it should be possible to just do:

```python
self.metadata["experiment_details"]  = self.get_params()
```

In my quick test it seems to produce the same output (although the keys are not in the same order) - if you think this works, it might be a bit cleaner because then the attacks don't need to explicitly define what to remove?

shahzadmumtaz22:
> This looks much cleaner, thanks!
> 
> If the following functions are included in the `Attack` class in attack.py, `get_params()` should return the parameters and their values.
> 
> ```python
> import inspect
> 
> 
>     @classmethod
>     def _get_param_names(cls):
>         """Get parameter names"""
>         init_signature = inspect.signature(cls.__init__)
>         parameters = [
>             p.name
>             for p in init_signature.parameters.values()
>             if p.name != "self" and p.kind != p.VAR_KEYWORD
>         ]
>         return parameters
> 
>     def get_params(self):
>         """
>         Get parameters for this attack.
> 
>         Returns
>         -------
>         params : dict
>             Parameter names mapped to their values.
>         """
>         out: dict = {}
>         for key in self._get_param_names():
>             out[key] = getattr(self, key)
>         return out
> ```
> 
> So instead of having to define the keys to exclude and calling `_exclude_keys_from_dict()` it should be possible to just do:
> 
> ```python
> self.metadata["experiment_details"]  = self.get_params()
> ```
> 
> In my quick test it seems to produce the same output (although the keys are not in the same order) - if you think this works, it might be a bit cleaner because then the attacks don't need to explicitly define what to remove?

Thanks @rpreen, for your feedback and tests. I have tested it through and it works fine so I updated it.

rpreen (approved):

rpreen:
Just wondering whether:

https://github.com/AI-SDC/AI-SDC/blob/48c820ab3c98bbb7d82e33c035b0ef0b87eff09a/aisdc/attacks/worst_case_attack.py#L52

Should be `RandomForestClassifier()` ?

shahzadmumtaz22:
> RandomForestClassifier

Later, we use this as shown. 

image

I wonder if should we then write RandomForestClassifier()?

rpreen:
Maybe @simonrnss should probably decide. I see how it's used there on line 328

https://github.com/AI-SDC/AI-SDC/blob/48c820ab3c98bbb7d82e33c035b0ef0b87eff09a/aisdc/attacks/worst_case_attack.py#L328

I guess it's just if I was a researcher looking at instantiating a worst case attack, I might assume that the parameter `mia_attack_model` was an instantiated model like `RandomForestClassifier()` or even `RandomForestClassifier(n_estimators=10, ...)` and in which case we wouldn't even need the parameter `mia_attack_model_hyp` because we could just get the values from the attack model with `mia_attack_model.get_params()` - and in that case the constructor default should be `mia_attack_model = RandomForestClassifier(min_samples_split=20, min_samples_leaf=10, max_depth=5)` It looks like at present the parameter value is a class and not an object, which obviously works currently, but is it the most user-friendly?

simonrnss (approved):

PR #179, opened 2023-06-20
Improve summary.txt generation module

yolaj-nhs:
Changes made:
- Do not overwrite repeated tests (currently each attack in the JSON file has an 'attack type', as a key e.g. AttributeInference, WorstCase, etc. The previous code meant that each time an attack of the same type was run, the results of the old attack were overridden. This has been fixed by appending the log_id of each test to each attack name, so it becomes WorstCaseAttack_[log_id])
- Added KNN to the list of models that are now immediately rejected
- Reformat the output of the JSON formatter so four items are now returned: the raw JSON output, a list of reasons for immediate model rejection (e.g. target is SVM), a list of reasons to support rejection, and a list of reasons to support release
- Changed the wording in the summary.txt file produced (see attached two files to see what these now look like)
- Add a function for allowing the code to take in and present results of target.json, as produced by PR #176 

[summary_safe.txt](https://github.com/AI-SDC/AI-SDC/files/11936643/summary_safe.txt)
[summary_svm.txt](https://github.com/AI-SDC/AI-SDC/files/11936644/summary_svm.txt)

Note: the next bit of work that needs doing is converting this file to be able to be run from the command line - but the PR already includes a lot of changes, so we'll leave it here for now!


aisdc/attacks/attack_report_formatter.py
jim-smith, line 1: @yolaj-nhs 1: you have a function that test for min-samples_leaf for rf, should we duplicate that for decision trees?
2: When looking at min_samples_leaf, please can we use the TREs coded risk appetite (currently in in AI-SDC/aisdc/safemodel/rules.json) rather than hard-coding values into the report.
3: There is a plausible argument that if we are working *after* the model has been created, then we should not be recommending rejection if a model that has the hyper-parameter 'min_samples_leaf' set to <threshold,. but we should be looking at the *actual* support for each region.   Trivial for DT but less so for RF. There is code that calculate 'k-anonymity' for a random forest, but trivially, we can find an upper bound on the membership of each  hyper-box in space but seeing how many records there are with the same output probabilities.

jim-smith: @yolaj-nhs This all seems fine. however, as a general point, there seem to be quite a few classes in attack_report_formatter.py where it seems the purpose of the instance is to be created , do one thing, and then die. And their names often sound more like function names instead of classes.
Is it worth considering jsut making these functions available within the scope of this file ?

jim-smith: @yolaj-nhs on reflection, ignore  point 3 above - it belongs as part of a wider discussion about measuring the regions of decision space that belongs in the next sprint I think

aisdc/attacks/attack_report_formatter.py
```diff
@@ -28,14 +28,31 @@ def __init__(self, filename=None):
...
+            else:
+                file_data = {}
+
+            f.truncate(0)
+
+        # Add the new JSON to the JSON that was in the file, and re-write
+        with open(self.filename, "w", encoding="utf-8") as f:
+            incoming_json = json.loads(incoming_json)
+
+            if "log_id" in incoming_json:
```
shahzadmumtaz22, line 51: @yolaj-nhs 

Thanks, overall this PR seems great to me. 

I wonder how it would be possible to have a log_id repeated. I believe the way we implemented it, each time the make_report(...) method is called, a unique UUID is supposed to be generated. Though I agree that storing output results from multiple attacks with the top-level JSON component as AttackName_LogID would be useful.

aisdc/attacks/attack_report_formatter.py
```diff
@@ -170,22 +232,37 @@ def __init__(self, report: dict, metrics_list=None):
...
+                            "max": max(metrics_dict[m]),
+                            "mean": np.mean(metrics_dict[m]),
+                            "median": np.median(metrics_dict[m]),
+                        }
+                    output_dict[k] = output
+        return (
+            output_dict,
+            self.immediate_rejection,
+            self.support_rejection,
+            self.support_release,
```
shahzadmumtaz22, line 264: In a class FinalRecommendationsModule, a method process_dict includes returning variables like immediate_rejections, support_rejection and support_release are updated and then returned.  Whereas in classes like  SummariseUnivariateMetricsModule, SummariseAUCPValsModule and LogLogROCModule, for process_dict method these variables are not updated and seem written just to maintain the returning variables pattern align with the process_diction method of the FinalRecommendationonly class. I wonder if could we think of some other way like when all these classes are inheriting a class AnalysisModule, should we not define these variables in the superclass and use them where required, and this will avoid repeating them across multiple classes?

tests/test_attack_report_formatter.py
```diff
@@ -209,13 +342,18 @@ def test_auc_pvals_module(self):
...
         json_formatted = self.get_test_report()
         f = SummariseFDIFPvalsModule(json_formatted)
         _ = str(f)
         _ = f.process_dict()
-        _ = f.get_metric_list(json_formatted["attack_experiment_logger"])
+        _ = f.get_metric_list(
+            json_formatted["WorstCaseAttack"]["attack_experiment_logger"]
+        )
 
         self.assertIn("Summary of FDIF p-values", str(f))
```
shahzadmumtaz22, line 359: Though I can see that you have updated the tests regarding the use of classes in attack_report_formatter module; however, I wonder if would it not be helpful to update the examples to demonstrate this functionality. 

I can see our examples are already cluttered in a way we are trying to demonstrate all possible scenarios, and we need to think about how can we make them clean. I believe this may be done by creating separate files to demonstrate running attacks from the command line and using programmatically for each type of attack.

aisdc/attacks/attack_report_formatter.py
yolaj-nhs: No that makes perfect sense - I spotted the same issue when I was writing the code, and I didn't like what I was writing, but I couldn't see a neat solution. This is exactly the solution I was looking for - thanks!

aisdc/attacks/attack_report_formatter.py
yolaj-nhs: A few notes on your comment:
- I will absolutely implement min_samples_leaf for decision trees
- I will use the risk appetite JSON file
- The support for each region (point 3 in your first comment) is trickier to do, I agree that it belongs in sprint 3
- Totally agree with being able to access each module separately, there's already a ticket for cleaning these modules which I will do first. Including that ticket in this PR would have caused a massive PR! After this has been merged, I'll start work on that ticket. Are you happy with this?

jim-smith: >Are you happy with this?
yep

yolaj-nhs:
From previous comments:
- [ ] min_samples_leaf for decision trees
- [ ] Risk appetite JSON file
- [ ] Update logging to uuid1?
- [ ] Refactor so recommend_rejection, support_rejection, support_release are in the superclass (as oppposed to being returned/passed around)
- [ ] Update examples

aisdc/attacks/attack_report_formatter.py
jim-smith: @yolaj-nhs don't think collisions are likely to be an issue. More pressing things to do :)

aisdc/attacks/attack_report_formatter.py
```diff
@@ -86,51 +128,86 @@ def __init__(self, report: dict):
...
+                self.immediate_rejection.append("Model is kNN")
                 return True
         return False
 
-    def _rf_min_samples_leaf(self, min_samples_leaf_score):
+    def _tree_min_samples_leaf(self, min_samples_leaf_score):
         if "model_params" in self.report:
             if "min_samples_leaf" in self.report["model_params"]:
                 min_samples_leaf = self.report["model_params"]["min_samples_leaf"]
                 if min_samples_leaf < 5:
```
jim-smith, line 149: line 149: 'if min_samples_leaf <5:" should be referring to the TRE risk appetite which you read earlier from file and stored in min_samples_leaf_score.   
(maybe you want to name that variable something else? like min_samples_leaf_appetite
so the line should be:    if min_samples_leaaf < min_samples_leaf_appetite:

aisdc/attacks/attack_report_formatter.py
```diff
@@ -86,51 +128,86 @@ def __init__(self, report: dict):
...
+                        if instance["P_HIGHER_AUC"] < p_val_thresh:
+                            stat_sig_auc.append(instance["AUC"])
+
+                    n_instances = len(
+                        self.report[k]["attack_experiment_logger"][
+                            "attack_instance_logger"
+                        ]
+                    )
+                    if (
+                        len(stat_sig_auc) / n_instances > 0.1
```
jim-smith, line 178: Fine for now, but for the next sprint suggest that we read these hard-coded values from a config fie or something so they can be changed on a per-dataset or per-tRE level.
Then I guess the output strings have to be more like; f"when we tested for the metric {metric_name} in more than {metric_tolerance} of our runs the value was statistically significant"

PR #180, opened 2023-06-21
add safemodel results to target.json and attack_results.json

rpreen:
In this PR:
* Safemodel SDC checks are written to `target.json` when calling `request_release()`
* With a target object, safemodel `request_release()`  runs all attacks and writes output to `attack_results.json`
* Example: `model.request_release(path="myTargetFolder", ext="pkl")` - This call will create a folder named `myTargetFolder` and it will contain `target.json` and `model.pkl` -- it will also include data such as `x_train.pkl` and the results of attacks in `attack_results.json` if a `target` object is supplied.
* Updated tests and checked that all examples still run.
* Updated notebooks to reflect current API.
* Run `python3 -m examples.safemodel_attack_integration_bothcalls` to see a full example.

Notes:
* The example scripts are starting to look pretty messy and could do with simplifying/cleaning (not part of this PR).
* The example notebooks also could use a clean and I'm not sure why we encourage researchers to call `save()` before calling `request_release()` when `request_release()` performs a model save anyway. Particularly since this PR operates on the basis that there will be one target model in each folder.
* PDF reports are no longer created with `request_release()` - only the json files. Once the improvements to the report generation have been completed it can automatically call that function to produce `summary.txt` etc.


aisdc/safemodel/safemodel.py
jim-smith, line 1: line 806 where we call the LIRA attack, number of shadow models is still hard-coded to 100.
do we need / can we avoid this?

tests/test_attacks_via_safemodel.py
jim-smith, line 1: Could add to the issue about improving tests the fact that these could do a better job of testing don't check correctness of output

jim-smith (approved):

PR #182, opened 2023-06-25
Run attack command line using target json and config json

shahzadmumtaz22:
- Enhanced functionality of running LIRA and Worstcase attack from the command line using saved target (target.json) and configuration files
- Examples updated
- Related tests updated


rpreen:
I think we need this for the attribute attack as well?

rpreen:
I wonder if we need a main entry point so one attack_config.json can fire multiple attacks?

shahzadmumtaz22:
> I think we need this for the attribute attack as well?

@rpreen 
Yes, I agree that we also need this for the attribute attack. I plan to make a separate PR for this.

shahzadmumtaz22:
> I wonder if we need a main entry point so one attack_config.json can fire multiple attacks?

@rpreen 

Yes, I have a plan to initiate a separate PR for a single entry point so we can run multiple attacks using a single configuration file and a target save model.

rpreen (approved):

PR #183, opened 2023-06-28
add generalisation error to target.json

rpreen:
Added as `test - train` - not sure if it should be a fraction?


jim-smith (approved):
Yeom et all express it as a simple difference

PR #188, opened 2023-06-30
AIA from the command line and standardised output format

shahzadmumtaz22:
A functionality to attribute inference attacks from the command line using configuration and target.json files.
A related test and example have been updated.


examples/attribute_inference_example.py
```diff
@@ -17,7 +21,7 @@
 
 from aisdc.attacks import attribute_attack  # pylint: disable = import-error
 from aisdc.attacks.attack_report_formatter import GenerateJSONModule
-from aisdc.attacks.target import Target
+from aisdc.attacks.target import Target  # pylint: disable = import-error
```
simonrnss, line 24: out of curiosity -- why don't we have an import error that needs to be disabled on line 23?

tests/test_attribute_inference_attack.py
```diff
@@ -95,7 +97,27 @@ def test_AIA_on_nursery():
...
+
+
+def test_AIA_on_nursery_from_cmd():
+    """tests running AIA on the nursery data
+    with an added continuous feature"""
+    target, _ = common_setup()
+    target.save(path="tests/test_aia_target")
+
+    config = {
+        "n_cpu": 7,
```
simonrnss, line 110: will 7 CPUs work in general -- need to make sure this script will run anywhere

simonrnss (approved):
Just a couple of tiny comments.

examples/attribute_inference_example.py
shahzadmumtaz22: I did check previously and it has some issues with linting. Anyways double checked this now and it was working fine so added disabling functionality.

tests/test_attribute_inference_attack.py
shahzadmumtaz22: Changed to 2 instead of 7; however, in case we don't pass this to AttributeAttack the default it takes as n_cpu: int = max(1, mp.cpu_count() - 1),

shahzadmumtaz22:
> Just a couple of tiny comments.

@simonrnss 

Addressed both comments and will merge PR later today.

rpreen (approved):
It looks ok; some minor thoughts:

* When running the example, it's annoying having the generated files spew all over the place - if not here then maybe we should make an issue to put them all in a single folder (same for all examples).
* A lot of the command line arguments seem overly verbose and could be shortened.

PR #194, opened 2023-07-05
Adding user stories

yolaj-nhs:
[user_stories_resources.zip](https://github.com/AI-SDC/AI-SDC/files/11958863/user_stories_resources.zip)

- [x] Use case 1, output produced:
[summary_user_story_1.txt](https://github.com/AI-SDC/AI-SDC/files/11958872/summary_user_story_1.txt)
- [ ] Use case 2
- [ ] Use case 3

### DIFF
--- a/example_notebooks/user_stories/user_story_1/user_story_1_researcher.py
+++ b/example_notebooks/user_stories/user_story_1/user_story_1_researcher.py
@@ -1,0 +1,125 @@
+"""
+User story 1 as researcher
+"""
+import logging
+import os
+import pandas as pd
+import numpy as np
+
+from sklearn.model_selection import train_test_split
+from sklearn.preprocessing import LabelEncoder, OneHotEncoder
+
+from aisdc.attacks.target import Target  # pylint: disable=import-error
+from aisdc.safemodel.classifiers import (
+    SafeDecisionTreeClassifier,  # pylint: disable=import-error
+)
+
+def main():
+    """
+    Create and train a model to be released
+    """
+    directory = "training_artefacts/"
+    print("Creating directory for training artefacts")
+
+    if not os.path.exists(directory):
+        os.makedirs(directory)
+
+    print()
+    print("Acting as researcher...")
+    print()
+
+    filename = 'user_stories_resources/dataset_26_nursery.csv'
+    print("Reading data from "+filename)
+    data = pd.read_csv(filename)
+
+    print()
+
+    y = np.asarray(data['class'])
+    x = np.asarray(data.drop(columns=['class'], inplace=False))
+
+    n_features = np.shape(x)[1]
+    indices: list[list[int]] = [
+        [0, 1, 2],  # parents
+        [3, 4, 5, 6, 7],  # has_nurs
+        [8, 9, 10, 11],  # form
+        [12, 13, 14, 15],  # children
+        [16, 17, 18],  # housing
+        [19, 20],  # finance
+        [21, 22, 23],  # social
+        [24, 25, 26],  # health
+    ]
+
+    # [Researcher] Split into training and test sets
+    # target model train / test split - these are strings
+    (
+        x_train_orig,
+        x_test_orig,
+        y_train_orig,
+        y_test_orig,
+    ) = train_test_split(
+        x,
+        y,
+        test_size=0.5,
+        stratify=y,
+        shuffle=True,
+    )
+
+    # [Researcher] Preprocess dataset
+    # one-hot encoding of features and integer encoding of labels
+    label_enc = LabelEncoder()
+    feature_enc = OneHotEncoder()
+    x_train = feature_enc.fit_transform(x_train_orig).toarray()
+    y_train = label_enc.fit_transform(y_train_orig)
+    x_test = feature_enc.transform(x_test_orig).toarray()
+    y_test = label_enc.transform(y_test_orig)
+
+    logging.getLogger('attack-reps').setLevel(logging.WARNING)
+    logging.getLogger('prep-attack-data').setLevel(logging.WARNING)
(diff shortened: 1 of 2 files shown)


example_notebooks/user_stories/user_story_1/user_story_1_researcher.py
```diff
@@ -0,0 +1,127 @@
...
+        target.add_feature(data.columns[i], indices[i], "onehot")
+
+    logging.info("Dataset: %s", target.name)
+    logging.info("Features: %s", target.features)
+    logging.info("x_train shape = %s", np.shape(target.x_train))
+    logging.info("y_train shape = %s", np.shape(target.y_train))
+    logging.info("x_test shape = %s", np.shape(target.x_test))
+    logging.info("y_test shape = %s", np.shape(target.y_test))
+
+    # [TRE / Researcher] Perform disclosure checks
```
jim-smith, line 101: maybe line 101 change the comment to:
# Researcher can check for themselves whether their model passes individual disclosure checks

example_notebooks/user_stories/user_story_1/user_story_1_researcher.py
```diff
@@ -0,0 +1,127 @@
...
+        logging.info("metadata is:")
+        for key, val in metadata.items():
+            if isinstance(val, dict):
+                logging.info(" %s ", key)
+                for key1, val2 in val.items():
+                    logging.info("  %s : %s", key1, val2)
+            else:
+                logging.info(" %s : %s", key, val)
+
+    # now via request_release()
```
jim-smith, line 119: maybe change comment in line 119 to
````# when researcher is satisfied the call request release()````
```` # if they pass in the target model object the cod will automatically run checks for the TRE staff````

File: example_notebooks/user_stories/user_story_2/user_story_2_best_case_researcher.py [hunk 3]
```diff
@@ -0,0 +1,121 @@
...
+
+    logging.getLogger("attack-reps").setLevel(logging.WARNING)
+    logging.getLogger("prep-attack-data").setLevel(logging.WARNING)
+    logging.getLogger("attack-from-preds").setLevel(logging.WARNING)
+
+    # Build a model
+    model = SafeDecisionTreeClassifier(random_state=1)
+    model.fit(x_train, y_train)
+    msg, disclosive = model.preliminary_check()
+
```
jim-smith, line 92: line 92 I think we should illustrate them calling model.request_release() but not giving it the target object.
In that case it will still produce the report on whether they did things like over-ride TRE-preferred values for hyper-parameters, change the models' apparent values for hyper-parameters  after calling the fit function, or change the actual content of the model (neural net, forest or tree) after calling fit (I.e. hide data in it)

jim-smith: continued: although I'm not sure that your generate report code would pick that up? If not needs adding as an issue for the next sprint

example_notebooks/user_stories/user_story_2/user_story_2_best_case_tre.py
```diff
@@ -0,0 +1,224 @@
...
+    # Wrap the model and data in a Target object
+    target = Target(model=target_model)
+    target.add_processed_data(x_train, y_train, x_test, y_test)
+    target.add_raw_data(x, y, x_train_orig, y_train_orig, x_test_orig, y_test_orig)
+    for i in range(n_features):
+        target.add_feature(data.columns[i], indices[i], "onehot")
+
+    SAVE_PATH = directory
+
+    # check direct method
```
jim-smith, line 94: if this is the script for the TRE, why do we need to call the attacks twice (once via the models' run_attack() method and once via request_release()?

Won't that just take extra time?

example_notebooks/user_stories/user_story_2/user_story_2_worst_case_researcher.py
```diff
@@ -0,0 +1,106 @@
...
+
+    logging.getLogger("attack-reps").setLevel(logging.WARNING)
+    logging.getLogger("prep-attack-data").setLevel(logging.WARNING)
+    logging.getLogger("attack-from-preds").setLevel(logging.WARNING)
+
+    # Build a model
+    model = SafeDecisionTreeClassifier(random_state=1)
+    model.fit(x_train, y_train)
+    _ = model.preliminary_check()
+
```
jim-smith, line 81: same comment as before about calling model.request_release() within params

example_notebooks/user_stories/user_story_2/user_story_2_worst_case_researcher.py
jim-smith, line 1: @yolaj-nhs is it confusing to have best and worst case in the same user journey(2)?

might make sense to split them out into two differently named user journeys?- just means making a different folder and moving files

jim-smith (changes requested):
just a few really minor changes to 
- improve wording of comments in a couple of researcher files
- show (in a couple of places) how researchers can call request_release() even if they don't hav the target object, and it will still run some 'good practice' checks on their model
- suggest splitting user journey 2 into two rather than having best/worst case scenarios.
-

example_notebooks/user_stories/user_story_2/user_story_2_best_case_tre.py
yolaj-nhs: That is a very good point, I was trying to show how to do both just to prove it can be done, but it doesn't add any benefit to this use case. I'll remove the top call - thanks!

jim-smith (approved):

simonrnss (approved):
Totally happy with this. Only slight change would be to add a little more context to the comments at the top of each file -- even if it's just a pointer to the location where the use cases are described more fully.

PR #200, opened 2023-07-10
Single entry run multiple attacks

shahzadmumtaz22:


aisdc/attacks/multiple_attacks.py
```diff
@@ -0,0 +1,203 @@
...
+        """
+        with open(self.config_filename, "r+", encoding="utf-8") as f:
+            file_contents = f.read()
+            if file_contents != "":
+                config_file_data = json.loads(file_contents)
+                for config_obj in config_file_data:
+                    params = config_file_data[config_obj]
+                    attack_name = config_obj.split("-")[0]
+                    attack_obj = None
+                    if attack_name == "worst_case":
```
yolaj-nhs, line 72: Tiny thing: to avoid code duplication, you could put the three attack_obj.attack(target) statements outside the if block, and call it once

aisdc/attacks/multiple_attacks.py
```diff
@@ -0,0 +1,203 @@
...
+                        attack_obj = WorstCaseAttack(**params)
+                        attack_obj.attack(target)
+                    elif attack_name == "lira":
+                        attack_obj = LIRAAttack(**params)
+                        attack_obj.attack(target)
+                    elif attack_name == "attribute":
```
yolaj-nhs, line 78: Can you add an else: return some kind of error statement, if the attack_name is not one of the three you test for here?

I can see a situation where we accidentally change the attack name and then get very confused when this fails in a strange way

aisdc/attacks/multiple_attacks.py
```diff
@@ -0,0 +1,203 @@
...
+                        attack_obj = AttributeAttack(**params)
+                        attack_obj.attack(target)
```
yolaj-nhs, line 80: This means that the config file won't close until the attacks are finished - right? Seems risky - could you move the if statements to outside the 'with open' block?

aisdc/attacks/multiple_attacks.py
```diff
@@ -0,0 +1,203 @@
...
+    def read_config_file(self) -> tuple[dict, int]:
+        """Reads a JSON configuration file and returns dictionary
+        with a number of configuration objects"""
+        with open(self.filename, encoding="utf-8") as f:
+            file_contents = f.read()
+            if file_contents != "":
+                config_file_data = json.loads(file_contents)
+            else:
+                config_file_data = {}
+        return config_file_data, len(config_file_data)
```
yolaj-nhs, line 131: Is there a reason this needs to return two parameters? I can see that the only place the second parameter is used is in the tests, when you check n == 7, but you could change the tests to len(config_file_data) == 7

Unless there's a reason it has to return both parameters obviously

yolaj-nhs (approved):
Some minor comments, approved after these have been resolved/answered

Thanks @yolaj-nhs for your feedback - Changes made as suggested.

aisdc/attacks/multiple_attacks.py
shahzadmumtaz22: Updated as suggested

aisdc/attacks/multiple_attacks.py
shahzadmumtaz22: Updated as suggested

aisdc/attacks/multiple_attacks.py
shahzadmumtaz22: We added seven attack configurations, so that is why testing to make sure that there are 7 attack configurations

yolaj-nhs (approved):
Looks good to me!

rpreen (approved):

PR #208, opened 2023-07-18
Spcifying output directory and raw json filename from example scripts

shahzadmumtaz22:
This PR focuses on putting all outputs (including pdf and JSON) files into a specified output folder. 
- On specifying the output directory, it generates a single pdf (the same way we have a single JSON output file) for each attack type unless you specify a different report_name value in each attack configuration - this is demonstrated in examples worst_case_attack_examle.py, likelihood_attack_example.py and attribute_inference_example.py.
- A special case of running multiple attacks using a single configuration file (which involves worstcase, lira and attribute) on specifying an output directory, by default, it creates three JSON outputs (one for each attack type) and three pdf outputs; however, in case to get a single JSON output and single pdf across attacks, one has to mention the same name (for report_name) in each attack configuration (demonstrated in MIAandAIA_attacks_example.py)
- Tests updated


rpreen (changes requested):
Needs the pypdf requirement added to `setup.py` and also the Sphinx `requirements.txt` which is in `docs/`

aisdc/attacks/attribute_attack.py
```diff
@@ -98,21 +107,29 @@ def make_report(self, json_attack_formatter=None) -> dict:
             Dictionary containing all attack output.
         """
         output = {}
-        logger.info("Starting report, report_name = %s", self.report_name)
+        report_dest = os.path.join(self.output_dir, self.report_name)
+        logger.info(
+            "Starting reports, pdf report name = %s, joson report name =%s",
```
rpreen, line 112: Typo `joson` should be `json` also missing space after the equals

shahzadmumtaz22: Sorry, I have fixed this now.

rpreen (approved):

PR #217, opened 2023-08-01
Updating user stories

yolaj-nhs:
Changes made:

- Move user stories folder to main folder
- Add user stories resources (dataset csv file) and update filepaths inside user story to match
- Add READMEs with user stories description to each user story
- Add James' R script for user story 4


rpreen:
Would a single `README.md` in the top level `user_stories` be easier for people rather than having to click through and read each one individually?

yolaj-nhs:
> Would a single `README.md` in the top level `user_stories` be easier for people rather than having to click through and read each one individually?

Yes - of course that would be easier, I don't know why I didn't do that originally!

rpreen:
I think some of these need some updating for the changes we made in v1.0.6. For example, in story 1: "User creates an object “mydata” of type aisdc.attacks.dataset.Data", but we don't have that class anymore, we have a Target class that holds both data and model information.

jim-smith (changes requested):
With the benefits of a couple of weeks away from this issue:

1. From a usability perspective, it would be fantastic to have a flow-chart associated with the user stories that researchers could see to tell TREs what story-line their work is on, and for TREs to understand what is going on.
2. From a TRE usability perspective, could we provide make your scripts inside each users story more generic - some of them only work for the specific researcher example we provide.. In other words provide a script that worked for any story in this path, then show an example of it being run for the specific example.
3. From the user-story 2 example (and maybe others) this would require the researcher proving the data pre-processing script in a separate file and the generic script asking for that location when it runs, alongside asking for location of indices and the dataset. Those things are hard-coded into the tre script - could we separate them please?
4. The readme is little confusing about what exists or doesn't - it says route 4 not supported but there is a folder with James' script in.
5. The instruction within the TRE scripts are not super helpful
 - we should aim to have the generic scripts available as part of acro because it is not reasonable to expect TRE operatives to know they have to cd /pytho_dir/lib/site-packages/aisdc/userstories
 - and also the paths suggested need updating anyway (since you moved user journeys out of notebooks)
 - and also the names are really long to remember and type: python -m example_notebooks.user_stories.user_story_1.user_story_1_tre
 - @yolaj-nhs @rpreen @simonrnss any preference for how we make these scripts easier to use?
- should we have a top level script that interactively navigated through flowchart to the different user stories ? (with shortcuts )
-

jim-smith:
@yolaj-nhs 
sorry, it has taken me ages to respond to this.

The flowchart is great - would be nice if the text was all in the boxes at some stage :)

Sorry to be a bore but 
I still think that it would be nicer and more TRE-friendly if they never had to edit python scripts, or to work out how to pass. the right arguments to a script manually.

Some output checkers would be put off by that, especially the "editing python " part  and worried about messing up the python inadvertently - for example by inconsistent use of space for tabs for indentation.

So my suggestion would be that rather than having anything hard-coded, or typing in parameters by hand, we do it via an editable config file.

In acro we use a .yaml file to hold the TRE's risk appetite because it is a more robust format.
Then in the acro constructor we read that into a config dictionary.e.g.
https://github.com/AI-SDC/ACRO/blob/main/acro/acro.py lines 61,64-76

the yaml file is at https://github.com/AI-SDC/ACRO/blob/main/acro/default.yaml

yolaj-nhs:
User Workflow
•	Researcher modifies the yaml file to include relevant parameters
•	The first entry in this file allows users to select which user story they should run – they can find out which using the attached flowchart
•	Users run the following command: generate_disclosure_risk_report.py
o	Note: they can provide a different .yaml file, the command then becomes: generate_disclosure_risk_report.py –config_file my_new_config_file.yaml
•	The TRE script of the relevant user story will be run, with the relevant files being put into a ‘release_files’ folder

user_stories/default_config.yaml
jim-smith, line 1: Could we add a couple of lines to the comments  making a clear distinction between:
- lines they need to change to post at **files they have already created** 
- lines they don't need to change (but can)  which are defining the names of files that **will be** created when the attacks run

simonrnss (changes requested):
Thanks @yolaj-nhs -- code all looks good to me. Some minor comments (can be progressed in this pr, or a new issue, or a combination)
1. user story 4 is not restricted to R. Should we make a python version too?
2. The R code could do with a trip through `lintr`
3. no TRE script for 4. Is that due to it being in R? A python story 4 could be sorted quickly I guess?

user_stories/user_story_1/user_story_1_researcher.py
jim-smith, line 1: Might be more helpful to TREs to understand, and for researchers to use as a template, if:

- the doc string at the top of the file  makes it clear that this is an example using a DT on the nursery data and doing some pre-processing before then running and testing a classifier
- we added some comments to split the code up into sections that did different things so researchers can see which they need to change (or not) if they want to just adapt this to a different problem and/or classifier
- we added some comments making it clear how they would adapt this for 
   -  different data
   - different preprocessing
   - a different model type

jim-smith: Also:
-  where we show the use of direct calls to attack - maybe add a comment to say that the researcher may the choose to change the model if it does not pass the tests
- where show the calls to request_release() maybe we should add a comment saying "you would not normally waste your and TRE time calling this unless you have already checked that your model is not disclosive or can provide a justification for an exception request"

user_stories/user_story_2/data_processing_researcher.py
jim-smith, line 1: Realyl ikje this, but it needs a doc string and more comments to show that this is an exemplar of what to do and how to comment it

user_stories/user_story_2/user_story_2_researcher.py
jim-smith, line 1: Again,
maybe the files doctoring could make it clear that this is one example of how a researcher might go through the process of in getting data, pre-processing it, then creating and assessing a target model.

jim-smith, line 1: Is it weird that we don't show the rtes`richer checking their model to see how accurate it is?

user_stories/user_story_2/user_story_2_tre.py
jim-smith, line 1: Maybe more detail in the doc string about what this file does
 - **and saying explicitly that the TRE should not need to edit it at all**

user_stories/user_story_3/user_story_3_researcher.py
jim-smith, line 1: Again,
everything is great bit it would be better as a training resource if there were more comments about what was going on and to split it up into sections of "get data", "pre-process data" "train model" "test model" "request release"

jim-smith, line 1: Doc string needs to make it clearer that the TRE should not have to edit the code.

Perhaps add a few more comments bout wha this code is doing please.

user_stories/user_story_7/user_story_7_researcher.py
jim-smith, line 1: Again, more comments to split the code into logical sections, so that it can function a a training resource as well pelase

user_stories/user_story_7/user_story_7_tre.py
jim-smith, line 1: Doc string could be more descriptive

jim-smith:
Pleas see above comments on individual files- it's all. about commenting

yolaj-nhs:
I still need to...
- Update the flow chart
- Update/create user story 4

jim-smith:
@yolaj-nhs thanks, much more self-documenting now.

jim-smith (approved):

jim-smith:
@simonrnss are you happy with this now and able to agree the PR? I haven't been through your suggestions.

yolaj-nhs:
To do prior to merging:

- Simon's approval
- Fix pre-commit hook
- Re-test user stories given recent changes in dev

simonrnss (approved):
All good to me, as long as the spelling error that the codecommit thing picked up is addressed...

PR #227, opened 2023-09-08
Removing a file that does not exist

yolaj-nhs:
This PR fixes a bug that occurs when aisdc can't run an attribute attack (because we haven't supplied enough data to the Target class), but the code still tries to remove the files that it expects to have been created

Note: this bug impacts the user stories so needs merged ASAP - thanks!


jim-smith (approved):
Looks good.
Surprised this was not picked up by our tests?

PR #229, opened 2023-09-15
Data update

albacrespi:
Fix broken links to data download.


yolaj-nhs:
Note: I'm still working on the code coverage/tests for this PR

rpreen:
I notice that the openml function always onehot encodes all of the features and always label encodes the target class; is that intended? wouldn't it depend on the specific dataset fetched from openml?

albacrespi:
Thanks for spotting this Richard, it was not intended. I think this function is better to be deleted for now as it won't work anyway.

jim-smith (approved):

jim-smith (changes requested):
Was the decision that this till needs to have commented out the one-hot encoding of categoric .features when loading oepn-ml datasets ?

jim-smith:
@albacrespi @rpreen 
Seems to add a set of experimental result files. Suggest these are removed and the results directory added to .gitignore so they don't get re-added

rpreen:
FYI there is no reason to continuously merge development back into a branch just because github says so.

rpreen (approved):

jim-smith (approved):

PR #234, opened 2023-10-09
fix package installation for python 3.8, 3.9, 3.10

rpreen:
Current status:
* Tests with Python38 **work** on all OSs - with the conditional setup.py included in this PR that installs older versions of tensorflow and tensorflow-privacy;
* Tests with Python39 **work** on all OSs;
* Tests with Python310 **work** on all OSs - after pinning statsmodels==0.13.5 (included in this PR) - necessary because of the new acro import in a recent PR;
* Tests with Python311 **fail** on all OSs because tensorflow-privacy will not work with 3.11 due to depending on pandas 1.4 and the related numpy version;

At this time, we can only support Python 3.8, 3.9 and 3.10 due to tensorflow-privacy.

It should be noted that some of the tests depend on the scikit-learn version that we use (currently 1.1.3) and break with newer versions.


jim-smith:
@rpreen agree that the tensorflow privacy things a real pain - I had to downgrade any tensorflow installation to 2.9  last week to get the tests to run properly.

It's making me really wonder whether we should be splitting this repo into two: aisdc for running attacks and doing risk analysis, and safemodels for the wrapper classes. 
Then the former need not have tensorflow at all. 
- unless we ever piggyback functionality in the safemodel classes during risk analysis ?

rpreen:
The macOS Python39 tests just picked up the following error (yet it was completing fine before) - is this assertion meant to be exact or should it be approximate?

```
=================================== FAILURES ===================================
_______________________ test_keras_model_compiled_as_DP ________________________

    def test_keras_model_compiled_as_DP():
        """Test Compile DP."""
        model, X, _, _, _ = make_small_model()
        loss = tf.keras.losses.CategoricalCrossentropy(
            from_logits=False, reduction=tf.losses.Reduction.NONE
        )
        model.compile(loss=loss, optimizer=None)
        isDP, _ = safekeras.check_optimizer_is_DP(model.optimizer)
        assert isDP, "failed check that optimizer is dP"
    
        right_epsilon = 20.363059561511612
        model.check_epsilon(X.shape[0], 20, 10)
>       assert (
            model.current_epsilon == right_epsilon
        ), "failed check for epsilon after compilation"
E       AssertionError: failed check for epsilon after compilation
E       assert 20.36305956151161 == 20.363059561511612
E        +  where 20.36305956151161 = <aisdc.safemodel.classifiers.safekeras.SafeKerasModel object at 0x13e9ace20>.current_epsilon

tests/test_safekeras2.py:[42](https://github.com/AI-SDC/AI-SDC/actions/runs/6468263015/job/17559935066#step:5:43)7: AssertionError
```

jim-smith:
@rpreen Yes, al; the checks for accuracy and epsilon values in test_safekeras2.py could be written to use:
``` assert actual = pytest.approx(desired,0.001) ```

Haven;t made the changes yet in case you have a version of the branch checked out?

rpreen:
Will leave that test change for another PR.

Am going to merge this because it's now critical to get the package working - CI tests confirm the status above.

PR #237, opened 2023-10-10
225 example notebooks should not change user paths but should run using installed acro

jim-smith:
Sorted. 
also put copies of the keras stuff into the development files folder in case we decide to remove them for now.


jim-smith:
can't adjust and test because aisdc doesn't currently install in clean venv on MacBook with M1 chip (tensorflow issue)

jim-smith:
Updated to remove all reference to local paths - works on fresh install of aisdcv.1.1 in a clean virtual environment

rpreen (approved):

PR #247, opened 2023-10-24
Increasing visibility of user stories + bug fix

yolaj-nhs:
Changes made:

- Fixed a bug related to rules.json path when running files from package
- Updated main README to add section on user stories
- Updated user stories README to make instructions clearer


albacrespi:
I think the text on the main page should go further up. I think it would make more sense before the section "content", but it could also be after or within, any thoughts?
Could you also add a link?

Once inside "user_stories":
- it is not clear to me why "default_config.yaml" is needed, and it might not be either for external users. Could you please add a sentence clarifying? 
- could you add subsections in the instructions?
- I wonder if adding a line of code of one example of how to run stories might be helpful? 
- the code on how to run scripts should be marked accordingly.

jim-smith (changes requested):
1. I think that Alba is right and the visibility of the user stories section should be moved higher up the readme file
2. I would like to reword the descriptions of the user stories. It is perhaps overly colloquial in places, and some of the the things it says are changed by, for example, the structural attacks.
3.

.gitignore
jim-smith, line 1: maybe also include names of folders of results produced by test scripts if these have not completed correctly ?

user_stories/user_story_3/README.MD
jim-smith, line 1: do we want to update this wording?

jim-smith (changes requested):
perhaps look again at readme for user story three?

albacrespi (commented):
@yolaj-nhs Can you deal with Jim's comment about names of folders please? The rest seems ok to be to be merged.

user_stories/default_config.yaml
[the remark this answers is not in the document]
albacrespi: done

albacrespi (approved):

jim-smith (approved):

PR #252, opened 2023-10-31
Main

jim-smith:


rpreen:
this is merging from main into development, which is the wrong way round

changes are supposed to go in development and then merged from development into main so it's streamlined

so this highlights again really why having a separate main branch is stupid because it creates unnecessary complications

rpreen (approved):

PR #257, opened 2024-02-01
add new datasets

albacrespi:
Modified loaders.py to include synth-ae (large, extra-large, XXL) and RDMP datasets. Removed Texas dataset.


tests/test_loaders.py
```diff
@@ -185,3 +198,25 @@ def test_synth_ae():
...
+    x_df, y_df = loaders.get_data_sklearn("synth-ae-extra-large", DATA_FOLDER)
+    assert x_df.shape == (8, 16), f"x_df shape is {x_df.shape}"
+    assert y_df.shape == (8, 1)
+
+    x_df, y_df = loaders.get_data_sklearn("synth-ae-XXL", DATA_FOLDER)
+    assert x_df.shape == (8, 16), f"x_df shape is {x_df.shape}"
+    assert y_df.shape == (8, 1)
+
+
+# def test_rdmp():
```
yolaj-nhs, line 215: The code coverage issue is because the RDMP dataset isn't being tested

I've written a test to cover this which will fix the code coverage issue, however the test is failing because the RDMP dataset isn't in the tests/datasets folder

Add the RDMP dataset into this folder, uncomment these lines, and the code coverage _should_ be fixed (I think - I may need to take another look to make sure but happy to do so)

yolaj-nhs (approved):

yolaj-nhs:
Approved - I've fixed the tests and the merge conflict, the code itself looks good to me

As mentioned above, the code coverage issue is because the RDMP dataset isn't being tested

I've written a test to cover this which will fix the code coverage issue, however the test is failing because the RDMP dataset isn't in the tests/datasets folder

If you want to improve code coverage, add the RDMP dataset into this folder, and uncomment the function at the bottom of the test_loaders.py file and the code coverage should be fixed. The tests might fail, but if so I'm happy to take another look, should be an easy fix

tests/test_loaders.py
jim-smith: @albacrespi please can we add the first five records from the RDMP data set to the test/datasets folder as Yola suggests.

rpreen (approved):

PR #259, opened 2024-02-21
Removing references to final score

yolaj-nhs:
Remove all references to a final recommendation in the user stories. Removed references in both code and example outputs. All associated tests pass.


tests/test_attack_report_formatter.py
```diff
@@ -28,7 +28,7 @@ def get_test_report():
     json_formatted = {
         "log_id": 1024,
         "metadata": {"attack": "WorstCase"},
-        "model_params": {"min_samples_leaf": 10},
+        "model_params": {"min_samples_leaf": 2},
```
simonrnss, line 31: what's this change for?

simonrnss (approved):
Just one small query. When merged we need to make sure Jim knows.

jim-smith (approved):

PR #262, opened 2024-04-12
update dependencies

### LINKED ISSUE #139: Package versioning and optional tf
AISDC requirements has very specific (and potentially problematic) package versions. This looks to cause problems on Jim's laptop (and probably elsewhere). It also takes a long time to install.

This is a placeholder for now.

## Jim's mac install issue

@simonrnss can test on a macbook pro
@rpreen can test on a macos runner

## Removing tensorflow

@rpreen to look into providing option to install AISDC without tf support as per openai gym

### LINKED ISSUE #253: Dependencies
Are the dependencies on Tensorflow and Tensorflow-privacy really required or can they be optional? It's giving issues to install on the TRE.

### LINKED ISSUE #258: Can the uuid package be removed from dependencies?
https://pypi.org/project/uuid/ is currently required:
https://github.com/AI-SDC/AI-SDC/blob/fa6c4d62f1274a18dd2044a51e1cf4b252146fd7/requirements.txt#L11

This package was last released Jun 20, 2006

Is it still needed? Is the built-in uuid library good enough? https://docs.python.org/3/library/uuid.html

rpreen:
This PR changes the installation requirements and dependencies:

* Removes support for Python 3.8, which is end of life;
* Adds support for Python 3.11;
* _Note:_ Python 3.12 cannot yet be supported because this seems to require the newest tensorflow 2.16, which doesn't seem to be compatible with tensorflow-privacy 0.9.0;
* `$ pip install aisdc` now installs **just** the attacks dependencies - this avoids any need to install the large tensorflow packages if only needing to run the attacks to assess privacy - all of the example scripts will run with this installation except the safemodel example;
* To fully install `aisdc`, i.e., to additionally install the safemodel dependencies run `$ pip install aisdc[safemodel]`
* tensorflow-privacy is updated to the latest 0.9.0; tensorflow is updated to 2.14; other dependencies are unrestricted;
* Removes the `requirements.txt` from the root as well as from `docs` to consolidate all version dependencies in one place  inside `setup.py` - this should help future changes because development will now install the local package with `$ pip install .`  - always do this inside a virtual environment;
* _Note:_ if someone wishes to run experiments and [freeze](https://pip.pypa.io/en/stable/cli/pip_freeze/) all versions for reproducibility, they should generate their own `requirements.txt` that will include the aisdc version; this should then be hosted in another repo along with experimental results;
* Updates the `README.md` to note the new installation steps;
* Refactors the tests such that all of the attacks tests are located in one subfolder and all of the safemodel tests are located in another; this enables running just the attacks tests with `$ pytest tests/attacks/` - the CI is still configured to install and test the whole of aisdc;
* Adds dependabot to automatically submit PRs to update github actions when necessary;


jim-smith (approved):
Looks good to me but I've added Simon and Alba as reviewers to make sure they re in the loop.
Does this have implications for Alba;s PR that adds extra datasets?

rpreen:
The modified `test_loaders.py` in the [dataset PR](https://github.com/AI-SDC/AI-SDC/pull/257) just needs to be moved into the new subfolder created here: `tests/preprocessing` after this is merged.

albacrespi (commented):
I see there are a number of changes in multiple places. Including Safe models, tests, attacks, tests etc. I see some of of them are just change of names and others updated code. And I understand some is new code for common tests?
The requirements, before there was a conditional statement, and now it is not, or I'm missing something? Also we had some comments from colleagues suggesting that some of the requerements could be made optional as it is only for some functionality of the tools, has the previous issues been resolved by including the new TensorFlow privacy?

rpreen:
> I see there are a number of changes in multiple places. Including Safe models, tests, attacks, tests etc. I see some of of them are just change of names and others updated code. And I understand some is new code for common tests? 

* There were some minor changes to the safemodel code to update it for the newer version of scikit-learn: basically just renaming `base_estimator` to `estimator`.
* Likewise, there was one change to the attacks code for the structural attack because it uses XGBoost and the newer version allows `n_estimators` to be `None`, which would cause the hyperparameter risk check to crash, so I simply set it to a large number (1000) in that event, which should never happen anyway assuming existing usage, which doesn't ever use a value of `None`.
* There have been no changes to the tests, except for a refactoring, which has split them into subfolders to separate attack code and safemodel code, etc.; as part of this I also removed some of the duplication by moving that code into a `common.py`.

> The requirements, before there was a conditional statement, and now it is not, or I'm missing something? Also we had some comments from colleagues suggesting that some of the requerements could be made optional as it is only for some functionality of the tools, has the previous issues been resolved by including the new TensorFlow privacy?

* The conditional statement in the original `requirements.txt` was to handle the special case of Python 3.8, which needs older versions of libraries like tensorflow and since 3.8 is due shortly for EOL, I have removed it here. I have also removed as many other dependencies as possible including the `uuid` library, which I think someone mentioned in the issues. It also pegs the dependency on `acro` to the newest version (used in structural attacks), which will determine some of the library versions used.

rpreen:
Can I suggest that from this PR onwards, we use the "squash and merge" option (select it from the arrow on the green "merge pull request" button)? This will keep the main commit history cleaner.

PR #269, opened 2024-04-24
remove useless pylint suppression pragmas

### LINKED ISSUE #264: Remove useless pylint suppression
There is a lot of useless pylint pragma suppression cluttering up the code.

If we enable `useless-suppression` in `.pylint.rc` it will flag them all so they can be removed.

rpreen:
Fixes #264

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -149,7 +149,6 @@ disable=raw-checker-failed,
         locally-disabled,
         file-ignored,
         suppressed-message,
-        useless-suppression,
         deprecated-pragma,
         use-symbolic-message-instead,
         invalid-name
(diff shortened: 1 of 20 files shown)


jim-smith (approved):
Great to be removing a load of unnecessary 'pylint ignores'.

Just a slight concern that some of them were put there in the first place not because pylint wanted them but because black or one of the other style checkers did?
Will the CI tests be picking them up?
If so happy for the merge to happen

rpreen:
It's only pylint that uses them and it's still set to run as before on the CI. The only reason we need the disables is because on the CI we set a pass/fail threshold of 10/10, which forces people to consider pylint's suggestions. Really the whole idea of cluttering up the code with pylint pragmas is horrible but at least it will now warn of useless ones too.

PR #281, opened 2024-06-04
add different implementation modes for online and offline LiRA

rpreen:
Changes:
* refactor LiRA attack;
* adds `mode: str` parameter to LiRA attack with options:
  - `"offline"` - our original offline norm cdf implementation (defaults to this);
  - `"offline-carlini"` - Carlini's offline logpdf implementation;
  - `"online-carlini"` - Carlini's online logpdf implementation;
* adds `fix_variance: bool` parameter to LiRA, which will use the global standard deviations instead of just for the individual record (defaults False);
* adds `n_normal` to the attack metrics - the fraction of (member + nonmember) records where $p\le0.05$ with a Shaprio-Wilk test for normality on the out samples;
* adds `report_individual: bool` parameter to LiRA to write individual record level results to the report JSON (defaults False);
  - individual report includes the $p$ from a Shaprio-Wilk test for normality on the out samples - if the test is unable to be performed because the variance is too small then the value is set to `np.NaN`;
  - If `report_individual=True` and `write_report=True` a plot of the individual LiRA scores is added to the PDF.


jim-smith (approved):
@rpreen this al looks good.
We need to raise an issue to generalise this to cope with cases where the target (and hence shadow classifiers) are not sklearn.base.estimators 

Can you confirm that the offline version gives the same results as Simon's original implementation?

rpreen:
Yes, running with the defaults (`mode="offline", fix_variance=False`) will produce the same results as before.

I think the random splitting needs to be looked at, but I have **not** changed it here, just broken things up into functions to make it easier to add the extra functionality mentioned above.

jim-smith:
@rpreen Cool, as long as you're confident that  'our' implementations work the same as before vey happy to approve this merge. thanks

PR #291, opened 2024-06-25
add CLI and tools for generating configs; significant refactor

rpreen:
tl;dr: existing code that uses `aisdc` will need some slight updating; see the examples folder README

*Note:* the user stories need updating or archiving to reflect these changes - for now a note has been added to the user stories README (and the user stories have been moved into the examples directory.)

Changes:

**Report Generation:**
* simplify report generation by adding `write_report: bool` to attack constructors which specifies whether to write JSON/PDF reports; calling `attack()` on an attack model generates and returns the report `dict`; eliminates the need for a subsequent call to `make_report()`;
* refactor attack report creation by moving generic code such as timestamps and ids to the `Attack` base class;
* change default output directories so that by default all outputs are written to the same location when running multiple attacks - this also involves removing the `report_name` parameter and hard coding it to "report" - the user specifies the folder to output not the specific JSON/PDF names;
* renamed `"experiment_details"` entry to `"attack_params"` and also renamed `"attack"` to `"attack_name"`  to be more descriptive;
* add `"target_model"` and `"target_model_params"` to the attack report;
* fix `WorstCaseAttack` PDF report where the glossary went off the page;
* fix removing the `roc_log.png` after running LiRA;
* JSON report now indented to make it more human readable;
* updated the documentation for these changes;

**Attacks:**
* refactor the attack constructors so that output directory handling and other generic parameters are dealt with in the `Attack` base class rather than duplicated in each subclass;
* simplified attack constructors, removed fail fast mechanisms;
* refactor `WorstCaseAttack` constructor:
   - `mia_attack_model` parameter renamed to: `attack_model` - also now is a string representing the class to instantiate to enable config file specification and also because the attack models are instantiated within the class anyway;
   - `mia_attack_model_hyp` renamed to `attack_model_params`
* add `factory.py` as a single source to create attack objects - used by config scripts and also safemodels;
* update logging calls to make output more understandable;
* prevent attempting to run an `AttributeAttack` if the features have not been defined;
* refactor `StructuralAttack` to split unnecessary risk and get parameter count into separate functions for each classifier; changed variables to `bool` type that were using 0 or 1 ints; renamed `DoF_risk` to `dof_risk`;

**Examples:**
* separate target model training into dedicated train scripts;
* rename programmatic attack examples and simplify to use the saved target models, which is a more likely scenario;
* move user stories into the examples directory

**CLI execution / config files:**
* remove existing config and CLI execution;
* target and attack configs now in yaml instead of json;
* generalisation error in `target.yaml` now either a `float` or `np.NaN`, previously was `str` when model wasn't fitted etc. `"data_name"` renamed `"dataset_name"`
* add main execution script to enable CLI running of attacks with config files, e.g., `$ aisdc run target_dir/ attack.yaml`
* add main execution script to enable CLI user-prompted generation of `attack.yaml` configs with: `$ aisdc gen-attack`
* add main execution script to enable CLI user-prompted generation of `target_dir/` with: `$ aisdc gen-target`
   - needs more work and feedback to handle different data/model sources - to be enhanced in a future PR

**General:**
 * update the README;
 * update version number to 1.2.0;

**Tests:**
* update tests to reflect above changes; refactor to simplify;


aisdc/config/utils.py
jim-smith, line 1: @rpreen is the chain of reasoning obvious enough for check_dir?  looks like the user has to either decide to remove a directory (in which case maybe they should be warned that this will delete the directory and all its previous contents) or if they say no, create a new one (in which case maybe we should say something like "Ok, please specify an alternative destination directory"

aisdc/config/target.py
jim-smith, line 1: @rpreen in _get_model_path() should we provide them with an exit route i.e. 'enter model path or Q to exit'
and then if they type 'Q' raise a ValueError to take them through to the next stage of asking for probabilities?

jim-smith, line 1: @rpreen actually in get_arrays() maybe we should offer them a get-out clause too?

Otherwise if you get half-way through the script and then type the wrong thing by accident you hast to ctrl-C out of the whole thing - and not every user may even be aware they can do that?

aisdc/config/attack.py
jim-smith, line 1: lines 48- 56 maybe need to give users an obvious way to get out of the while True loop  I.e. enter X to exit this step

jim-smith, line 1: line 76 maybe say 'Get Default config (will run all attacks)?'

examples/README.md
jim-smith, line 1: @rpreen Maybe at least once expand the acronym CLI :)

examples/train_rf_nursery.py
jim-smith, line 1: a couple more comments about how the indices directory is constructed and what the point is might be useful?
I.e. saying how many values each  raw categorical feature has or something

tests/attacks/test_lira_attack.py
jim-smith, line 1: are we losing a set of assertions/tests about what the lira attack should actually be doing?

aisdc/config/attack.py
rpreen: lets refine the config tools in another PR, this gets a basic working version in place

tests/attacks/test_lira_attack.py
rpreen: There weren't very many assertions in tests to begin with, so should be improved in another PR

CHANGELOG.md
jim-smith, line 1: @rpreen I think this needs to state exactly that
-  the output report has moved to yaml to json in case there are people out there who are set for ingesting json. 
- if users want to change params from defaults they now need to do this via arguments to call rather than via a config file

rpreen: the report output has not changed from JSON since that will likely be ingested by another program; but when saving/loading a target object which contains paths to the data etc. that has changed from JSON to YAML to better enable hand modified versions for example if the model was not produced in Python or with any of the tools; and the CLI that uses a config file to specify the attacks to use is now YAML again because it's likely that a human will want to read or modify it

aisdc/config/attack.py
jim-smith, line 1: @rpreen so we don't lose them, please raise a new issue to resolve some of the comments later e.g. providing a way of backing out without hitting ctrl-c

rpreen: added #294

aisdc/config/target.py
jim-smith, line 1: @rpreen line 65 should we remove hard-coded int stipulating a max of 64 features and instead use a constant and definer that somewhere else?
(I imagine num features could get huge for genomics data)

rpreen: do you really think someone is going to manually sit there and add more than 64 features in a CLI prompt?

rpreen: I just added a constant and updated the display message to let the user know they can manually add them to the `target.yaml` after

jim-smith (approved):

jim-smith:
@rpreen all checked and seems fine.
Happy for you to merge.
Can't remember why I had the structural attacks returning ints but your rewrite is much cleaner :)

Issue #294, opened 2024-07-04
Improve/extend CLI prompt tools


rpreen:
As of `1.2.0` the CLI prompt tools were added: 
* `$ sacroml gen-attack` tool that produces the attack config  `attack.yaml` for running attacks on the CLI
* `$ sacroml gen-target` tool that produces the directory `target_dir/` that contains a `target.yaml` and model/data files for running attacks on the CLI
* subsequently running `$ sacroml run target_dir/ attack.yaml` executes the attacks and generates a report

These tools need to be improved and extended to provide a better user experience, such as more graceful exiting, and support for different scenarios when the model was not trained in Python, and where the data may be large and not able to be copied and also to handle various preprocessing situations. It seems likely that the `sacroml/attacks/target.py` will also need to be modified and an `sacroml/attacks/model.py` may need to be created to wrap target models to encapsulate non-scikit-learn models with `fit()`, `predict()` etc. functions that the attacks can assume exists.

rpreen:
At the moment, when you `sacroml run` with a given target folder and attack config, it tries to load the saved model and data files from the target folder into a `Target` object with `load()` in target.py. If everything is loaded in the `Target` object everything can run as normal.

https://github.com/AI-SDC/SACRO-ML/blob/c7d3ec39a0ae1d092b5386d9aab03cd1bb025b48/sacroml/attacks/target.py#L350-L389

If a `model_path` is specified in the target.yaml (in the target folder) it will try to load the model, but it only supports scikit-learn models currently and will throw an exception if it's not a pkl model. 

The worst case attack checks if there is a loaded model and training data available and then generates the probabilities as usual; if that's not available, then it checks if the predicted probabilities are available and uses those (as in user story 4). 

https://github.com/AI-SDC/SACRO-ML/blob/c7d3ec39a0ae1d092b5386d9aab03cd1bb025b48/sacroml/attacks/worst_case_attack.py#L116-L134

**TODO:** 

- [x] Update target.py so that the train/test probabilities can be loaded from the target.yaml.
- [ ] Update target.py to support loading models other than just *.pkl.
- [x] Add better handling of unsupported models to the attack modules.
- [x] Copy unsupported models to the target directory
- [x] Deal with situations where the datasets are large (currently they get copied to the target dir)
- [x] Deal with complex preprocessing scenarios

rpreen:
[...]

PR #323, opened 2025-02-10
add save/load target train/test probas

rpreen:
Partly addresses #294 

* Adds saving/loading Target model training and testing probability arrays to help support user story 4 from the CLI.
* Adds Target object loading csv files as well as pkl
* Adds additional checks that attacks can be run and skips them (with a logger warning) instead of raising exceptions

Supports user story 4 where `sacroml gen-target` is run and any model is provided along with csv files (csv files should not have headers) for probas and processed data (X_train, etc.), Subsequently `sacroml gen-attack` with defaults and then running `sacroml run target/ attack.yaml` In this scenario only the worst case attack can run because the others require a loaded model. Note it also wont yet copy the model to the target directory unless it's a scikit-learn model.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Changes:
 *   Fix AdaBoostClassifier structural test ([#318](https://github.com/AI-SDC/SACRO-ML/pull/318))
 *   Fix user story examples ([#320](https://github.com/AI-SDC/SACRO-ML/pull/320), [#321](https://github.com/AI-SDC/SACRO-ML/pull/321))
 *   Update acro dependency to v0.4.8 ([#322](https://github.com/AI-SDC/SACRO-ML/pull/322))
+*   Add Target loading of probas from persistent storage ([#323](https://github.com/AI-SDC/SACRO-ML/pull/323))
 
 ## Version 1.2.1 (Jul 29, 2024)
 
--- a/sacroml/attacks/attribute_attack.py
+++ b/sacroml/attacks/attribute_attack.py
@@ -66,7 +66,7 @@ class AttributeAttack(Attack):
             Attack report.
         """
         if target.n_features < 1:
-            logger.info("Can't run attribute inference unless features are defined.")
+            logger.info("WARNING: AttributeAttack requires features to be defined.")
         else:
             logger.info("Running attribute inference attack")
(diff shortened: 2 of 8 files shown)


jim-smith (approved):
These all seem sensible steps on the pathway to wider applicability

albacrespi (approved):

albacrespi:
When I run `sacroml gen-target` and `sacroml gen-attack` it feels like the first checking of paths etc is a bit slow, before any question is prompt. Is that expected behaviour or is it something to do with machine?

Is it expected behaviour when I create the target doesn’t seem to parse/write down some answers to the target.yaml file. It does write down the name of the dataset but none of the other parameters. Is it because of the format of the files? Also not that the format of the model is not accepted either.

> INFO:numexpr.utils:NumExpr defaulting to 4 threads. 
> Do you have a saved fitted model? (yes/no): yes
> Enter path including the full filename: /Users/a/SACRO/model.rds 
> Unable to load model.                                                               
> Do you have the model predicted probabilities? (yes/no): yes 
> What is the path to proba_train? /Users/a/SACRO/output_train.csv                                                       
> WARNING: Target cannot load .csv files.                                             
> What is the path to proba_test? /Users/a/SACRO/output_test.csv                                                         
> WARNING: Target cannot load .csv files.                                             
> What is the name of the dataset? user_story4                                        
> Do you know the paths to processed data? (yes/no): no                              
> Do you know the paths to original raw data? (yes/no): no                 
> Target generated in directory: 'target'

The content of the target.yaml file is:

> dataset_name: user_story4                                                          
> n_samples: 0                                                                        
> features: {}                                                                       
> n_features: 0                                                                     
> n_samples_orig: 0                                                                  
> generalisation_error: .nan                                                         
> safemodel: []                                                                      
> X_train_path: ''                                                                  
> y_train_path: ''                                                                  
> X_test_path: ''                                                                  
> y_test_path: ''                                                                    
> X_orig_path: ''                                                                    
> y_orig_path: ''                                                                    
> X_train_orig_path: ''                                                              
> y_train_orig_path: ''                                                              
> X_test_orig_path: ''                                                               
> y_test_orig_path: ''

rpreen:
```
WARNING: Target cannot load .csv files.
```

Looks like it's not using the updated version - you might need to reinstall the package from your local branch with `python -m pip install .` from the root directory

PR #330, opened 2025-03-25
abstract target model and enable support for Pytorch

### LINKED ISSUE #236: Add support for PyTorch
Maybe even phase out safekeras since tensorflow privacy is so dependent on out of data packages

rpreen:
Changes:

* Add an abstract target `Model` class for hiding model-specific implementation details from attacks, etc.;
    - Modify attack code calls to target model to use this abstraction;
    - Add `attackable()` function to each attack implementation to check a `Target` object is suitable for the attack;
* Add `SklearnModel` class that implements the existing target model functionality for scikit-learn models;
    - Existing examples using scikit-learn should function without any changes needed;
* Add `PytorchModel` class for supporting pytorch - Resolves #236 
    - Seems to be working with all of the attacks: the `StructuralAttack` will be skipped since it requires target models of `sklearn.base.BaseEstimator`
    - Implements generic functions like `score()` `clone()` `predict_proba()` and `predict()` - returns integer labels;
    - Researchers must provide a model class and a train function - see `examples/pytorch/`
        + The `train` function currently requires two numpy arrays as inputs - we need to move to using `DataLoader` 
classes for all models later;
    - **GPU support:** seems to work with limited testing; **usage needs further testing;** note that there is still a lot of copying back to the CPU when running attacks currently;
* Add `Target` object support for saving (and loading) the researcher's Python code to the target directory as well as parameters needed for instantiating the model and executing training;
* Add Pytorch example to examples folder.
* Modify the dependencies in `pyproject.toml`
    - Add torch to base requirements; (for pytorch)
    - Add prompt-toolkit to base requirements; (for CLI tab completion)
* Update `AttributeAttack` to fallback to sequential processing instead of parallel for models that are not `sklearn.base.BaseEstimator` to enable pytorch models to work - for some reason multiprocessing pickles models and doesn't seem to immediately work with pytorch;
* Update `LIRAAttack` to cope with both one-hot encoding as well as integer label encoding (existing);
 * Update tests;
    -  Add test module for `PytorchModel` to `tests/attacks/`
* Update CLI `gen-target` prompt to include Pytorch files/params - **needs additional testing**
* Add tab completion to file paths when using CLI prompts
 * Incremented the version number to 1.3.0


examples/pytorch.py
jim-smith, line 1: 1. Making the example easier ot repurpose
- maybe define a function that returns the dataset to make the code in the main function simpler?
- maybe some comments  like ```#### your code to define model below here ######``` and ```#### your code to define model above here ######```` ?

2. Fixed training vs early stopping
-Do we need to insist that people have fixed number of epochs?  Presumably this is for when we create shadow models in LIRA but it seems a little restrictive
- or could it be max epochs and if so what do we need to know to support early stopping?

sacroml/attacks/model.py
jim-smith, line 1: There's an implicit assumption here that target models are classifiers. I wonder if we should make this more upfront?
We are being asked advise on all sorts of things - like attention transformswers and other foundation models. I know that these would not be attacked using these attribute/membership inference attacks per-se but where do we document that? and what is the expected behaviour is someone runs attacks on a model that has been to do a regression - or even a multiu-label classification problem?

sacroml/attacks/model_sklearn.py
jim-smith, line 1: same comment about setters return changed model rathethan doing it inplace

also here and in PyTorch model should the generalisation error method make it clear in the docstrings that  it is expecting preprocessed data?

tests/attacks/test_pytorch.py
jim-smith, line 1: do the tests not use a standard dataset that is kept in memory somehow? (as a fixture?)

sacroml/attacks/model_sklearn.py
rpreen: The setters are changing the parameters in place; it returns `self` for sklearn method chaining.

The generalisation error function is only to be called internally from the Target object, which controls what data is passed - researchers wont use it.

tests/attacks/test_pytorch.py
rpreen: Tests are a mess and use a variety of different dataset functions still (I did some refactoring before which has reduced it a little) - they need a massive refactor, but not in this PR.

sacroml/attacks/model.py
rpreen: The whole package assumes classifiers at this stage, so that should be made clear in the main README?

examples/pytorch.py
rpreen: I have already started splitting it apart into `model.py`, `dataset.py`, `train_pytorch.py` and `attack_pytorch.py` - the model and dataset code/modules need to be copied to the target_dir/ so they can be loaded later.

The attributes such as `epochs` are there mainly to automatically write the hyperparameters to the target.yaml, what goes in the fit function could and should be anything so would allow early stopping, etc. - may be necessary to drop the idea of trying to write a description of the hyperparameters so those attributes don't have to be defined. Currently only the `layers` attribute is needed in `PytorchModel` -

sacroml/config/target.py
jim-smith, line 1: line 129- maybe say what you want the path to? - other prompts are more explanatory

jim-smith:
one comment to address then happy to approve.

sacroml/config/target.py
rpreen: Done

jim-smith (approved):

PR #335, opened 2025-04-17
remove tensorflow-privacy support including safekeras/safetf

### LINKED ISSUE #334: [Behaviour Change] Remove support for tensorflow-privacy
* tensorflow is not available in conda for Windows;
* tensorflow-privacy is a major pain limiting sacroml to Python <3.12

Removing tensorflow and tensorflow-privacy will make package installation easier, but it will require the removal of safekeras and safetf - these could be readded later.

rpreen:
Resolves #334 

CI [confirms](https://github.com/AI-SDC/SACRO-ML/actions/runs/14516432628) that Python 3.9 -- 3.13 on macOS, Windows, and Ubuntu are all installing and running tests successfully with these changes.


jim-smith:
@rpreen assuming this still keeps dictdiffer?

rpreen:
Yes, it keeps dictdiffer.

rpreen:
Attacks should still work on a keras model - or any model - if it inherits from `sklearn.base.BaseEstimator` and it's pretty easy to create a wrapper to do so.

In the pytorch branch, I currently have an _incomplete_ wrapper for keras, but for now this will allow me to drop that and keep tensorflow out. We can just make an issue and deal with keras/tensorflow when someone really wants it? 

Possibly could complete that keras wrapper (in another PR) similar to the pytorch one and make it dynamically load dependencies if someone manually chooses to install keras/tensorflow so the wrapper code is already present; would need to see if that's simple or messy.

jim-smith:
@rpreen maybe rebrand as remove support for tensorflow-privacy

jim-smith (approved):

jim-smith (approved):

PR #345, opened 2025-06-29
improve data handling for users

rpreen:
Improve data handling for users:
* Add abstract class `sacroml.attacks.data.PyTorchDataHandler` that defines a standard interface for PyTorch dataset handling;
    - This contains a constructor, `get_dataset()`, `get_raw_dataset()` and `get_dataloader()` functions;
* Update `sacroml.attacks.target.Target` data wrapping:
    - Option to pass numpy arrays directly as `X_train`, `y_train`, etc., still exists - these are copied to `target_dir/`
    - Alternative option is now to pass dataset code to `dataset_module_path`, along with the class name in `dataset_name` and the train/test indices as `indices_train` and `indices_test` - in this case, the code and indices are copied to the `target_dir/` - the indices must be supplied because random seeds cannot be relied upon;
    - Because the attacks still heavily depend on numpy arrays, upon loading a `target_dir/` the data is automatically converted to numpy (for example to perform data shuffling for LiRA) - this is then converted back to dataloaders so that the shadow models are still trained in the original way, etc. This means this PR can focus on improving the user interface without modifying existing attack code. This should be improved in a future PR - but it seems the only major drawback is that the dataset must still fit in RAM and scalability issues need to be addressed separately.
* Add abstract class `sacroml.attacks.data.SklearnDataHandler` that defines a standard interface for scikit-learn dataset handling;
    - This contains a constructor, `get_data()`, `get_raw_data()` and `get_subset()` functions;
    - This can be used similarly to the PyTorch dataset above;
* Update / refactor example(s):
    - There is now a Simple example and CIFAR10 example - see `examples/pytorch`
    - The `examples/sklearn/cancer` example shows original use passing numpy arrays directly;
    - The `examples/sklearn/nursery` example shows how to use a supplied dataset module;
* Update tests to reflect changes and address various code coverage issues;
* Update `$ gen-target` to prompt for a dataset module and indices;
    - Extract class names from supplied `model.py` and `dataset.py` to use for autocompletion.


CHANGELOG.md
jim-smith, line 1: @rpreen maybe make explicit this new supports data held in different files within directories rather than just in a single file

examples/pytorch/cifar/dataset.py
jim-smith, line 1: A few more comments  would help explain the example ---making it clearer (maybe with URLs to PyTorch examples) which parts are standard PyTorch dataloader bits and which are new/specific to sacro-ml

- That might make it clearer how to use this as a template that could be adapted to a new task.   
**Or**  maybe we c/should provide a template?

examples/pytorch/cifar/model.py
jim-smith, line 1: typo? kernal->kernel

examples/pytorch/cifar/train.py
jim-smith, line 1: If this is all completely standard code that you've taken from an online example then say so at the top of the file and maybe even ref the source.

examples/pytorch/cifar/train_pytorch.py
jim-smith, line 1: Change docstring to Train a PyTorch classifier on CIFAR10 using the sacroml Target and dataset classes.

Make it clearer which bits are standard PyTorch stuff and which is stuff needed to create an sacroml target object.

examples/pytorch/simple/attack_pytorch.py
jim-smith, line 1: change docstring to say this is an example of how to run attacks 
- on a simple PyTorch model that has been saved as an object of type target. 
- with datasets that are in a single file (if that is relevant)

examples/pytorch/simple/dataset.py
jim-smith, line 1: again could be lots more information in the docstring and in subsequent comments saying how this could be adapted - especially a comment before line21 to make it explicit we are creating a dataset with 50 samples, 4 features and 4 classes

examples/pytorch/simple/train.py
jim-smith, line 1: Again add to docstring saying there is everything  in here is standard PyTorch stuff , jsut broken into functions the way we need it

examples/sklearn/README.md
jim-smith, line 1: maybe add comment saying there is no particular reason for doing cancer one way and nursery the other. Either could have been done either way

examples/sklearn/cancer/attack_rf_cancer.py
jim-smith, line 1: maybe add a comment saying why you catch the output from the first set of attacks but then over-write it (rather than append) with the second set?

examples/sklearn/nursery/dataset.py
jim-smith, line 1: add to docstring to say : "Nursery dataset handler for scikit-learn models illustrating the use of the SklearnDataHandler class from sacroml"

Also add comments to make more explicit which of these methods would need to be changed when adapting for a new dataset and which can be simply copied and left alone

examples/sklearn/nursery/train_rf_nursery.py
jim-smith, line 1: Again add to docstring saying what features this is illustrating

I know you have the line 
```logging.info("Wrapping the model and data in a Target object")```

but maybe a comment line might make it explicit this is the bit needed for sacroml

sacroml/attacks/model_pytorch.py
jim-smith, line 1: This is quite dry and a lot to read. I guess the idea is that people don't ever need look in here
- the rst docfiles that get generated should be all that is needed
- and the examples also show how it should be done.
- do the .rst files get generated before the final merge is done?

tests/attacks/pytorch_dataset.py
jim-smith, line 1: Am I right in thinking this file is a duplicate of something from the examples directory?  Does it matter if they ever get out of step?

tests/attacks/test_pytorch.py
jim-smith, line 1: feels like the tests are making sure that things (data, outputs) are created but not what they are.
could you test the functionality of the code for PyTorch data classes that turns things back into numpy arrays in a more exact way? (test for equality with the original data you moade)
Maybe it is worth adding a comment that the **functionality** of the attacks etc. (I.e. the contents of the outputs created) is covered by a different set of tests

tests/attacks/test_sklearn.py
jim-smith, line 1: same comments that the assertions mostly test things are not null rather than being more precise. 
Maybe it is worth adding a comment that the functionality of the attacks etc. is covered by a different set of tests

jim-smith (commented):
lots of minor requests for comments but nothing particularly in the code itself

sacroml/attacks/model_pytorch.py
rpreen: Researchers should never need to know anything about the abstract model classes. So it is intentionally omitted from the API documentation to not cause confusion or clutter.

tests/attacks/pytorch_dataset.py
rpreen: Yes, it's a duplicate. Necessary since one is a test and the other is example. And yes they will need to be be kept in synch if there are changes.

tests/attacks/test_pytorch.py
rpreen: Yes, tests need massively improving, but that goes for the whole project. We should do that in another PR otherwise it's going to block this from being merged for a while. Lets just get this merged and a new release pushed because there are quite a few changes to the examples to make it easier for new users and the package dependencies need to be updated due to the scipy issue. We can then iterate.

PR #362, opened 2025-07-30
Refactor and doc structural attack

jim-smith:


jim-smith (commented):
a few minor changes to make

sacroml/attacks/structural_attack.py
jim-smith, line 1: use Notes  and references syntax within docstring say the methodology  for the unecessary risk criteria is described in the aXiv paper https://doi.org/10.48550/arXiv.2502.09396

use notes/refs  in docstring for _get_unnecessary_risk_xgb to give links to xgboost repos where the default ar taken from

docstring for _assess_class_disclosure_risk is not quite right.  It assess whether privacy is leaked because the model reports that all members of a a subgroup do/don't belong to a given class

jim-smith:
delete the lines that tests that can never reach (351-352, 361-362) 
because they both   tests whether equiv_counts.size==0   
but that can never happen becuase equiv_counts comes from a call to np.unique(marry, return_counts=True)     
and the counts could only be an empty list if there were no unique values
- which is logically impossible = even  np.unique on an array of size n where every record is the same value a, returns (a,n)

jim-smith:
if you wrote a test within test_structural_attack.py that created an attack object -- but did not run the attack,  but then tried ot get the report that should trigger the missing line of code at the end

PR #367, opened 2025-08-27
Add record level reporting and PyTorch support to StructuralAttack

### LINKED ISSUE #364: [New Feature Request] Extend structural attacks : PyTorch models and store record-level outcomes
Structural attacks don't currently provide the option to store record-level outcomes  
Structural attacks should support PyTorch models

As above
for records store k-anonymity and class disclosure plus model level things (DoF)
support ingestion of PyTorch models, and extraction of the number of trainable parameters they have
start off with sklearn MLPs?
- going to need PyTorch for CNNs anyway

**Additional context**

kayatefi:
* Add store record level 
* Support PyTorch models

### DIFF
--- a/sacroml/attacks/structural_attack.py
+++ b/sacroml/attacks/structural_attack.py
@@ -30,6 +30,11 @@ from sklearn.neural_network import MLPClassifier
 from sklearn.tree import DecisionTreeClassifier
 from xgboost.sklearn import XGBClassifier
 
+try:
+    import torch
+except ImportError:
+    torch = None
+
 from sacroml.attacks import report
 from sacroml.attacks.attack import Attack
 from sacroml.attacks.target import Target
@@ -39,6 +44,15 @@ logger = logging.getLogger(__name__)
 
 
 # --- Data Structure for Attack Results ---
+
+
+@dataclass
+class StructuralRecordLevelResults:
+    """Dataclass to store record-level outcomes for structural attack."""
+
+    k_anonymity: list[int]
+    class_disclosure: list[bool]
+    dof: list[int]
 
 
 @dataclass
@@ -203,6 +217,8 @@ def get_model_param_count(model: BaseEstimator) -> int:
     -------
     int : Estimated number of learned parameters.
     """
+    if torch is not None and isinstance(model, torch.nn.Module):
+        return sum(p.numel() for p in model.parameters() if p.requires_grad)
     if isinstance(model, DecisionTreeClassifier):
         return _get_model_param_count_dt(model)
     if isinstance(model, RandomForestClassifier):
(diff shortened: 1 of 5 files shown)


sacroml/attacks/structural_attack.py
jim-smith, line 1: line 419 is a test for what I have called 'lowvals_class disclsoure' I.e. where model is effectively saying 'for records in this space a small group are of one class, and the size of that group is smaller than we would normally report on'

should also have a test for 'pure' class disclosure I.e. where the class_probs==0 - does this apply other than the main structural attack code as well?

Also are lines 415-420 creating one entry per sample, or one per 'equivalence class?

jim-smith, line 1: line 346-350 function attackable()
should it have a clause allowing it to attack pytorch models I.e.

```` if (
            target.has_model()
            and isinstance(target.model.model, BaseEstimator)
            and target.has_data()
        ): 
```` 
should be:

````
if (
            target.has_model()
            and (isinstance(target.model.model, BaseEstimator) or isInstance (target.model.model,PyTocthModel))
            and target.has_data()
        ):
````
- I'll need to check what  our new base class is called for PyTorch models and amend**
- maybe it would be cleaner if we should have a list of supported types at the top and then just loop over that?


Also as discussed now we have this I can look at these lines which define class disclosure and how we store record-level k-anonymity:

````
       # Populate record-level k-anonymity and class disclosure
        for count, class_probs in zip(equiv_counts, equiv_classes):
            self.record_level_results["k_anonymity"].append(count)
            self.record_level_results["class_disclosure"].append(
                bool(np.any((class_probs > 0) & (class_probs < self.THRESHOLD)))
            )

````

File: tests/attacks/test_structural_attack.py
Comment (jim-smith) [inline tests/attacks/test_structural_attack.py:1]: line 412 tests the algorithm that counts PyTorch parameters gives the same on its own as when called from the structural attack on a model.

But you;re not checking that it is the value you expect (65).  What if both are returning zero?
- would be good to add one more assert statement

**Comment by jim-smith** [comment]
@kayatefi Before we can merge this pull request we should define a simple case where we know what the correct record-level output should be and then write an appropriate test to see it is generated properly.

I was thinking of doing something simple with a decision tree and a small dataset- maybe 20 cases (some being duplicates), of 2/3 variables with an xor decision function.
- scenario 1: decision tree max depth 1: should get 50% accuracy, at least 10 records in each leaf so 5 of each class ==> should fail all the strucutural tests
- scenario 2: let the tree grow deeper: should her able to manipulate it to show class disclosure and small group problems

**Comment by jim-smith** [comment]
@rpreen I've been through Kay's code which was mostly fine then added the individual record-level reporting, which meant I had to rewrite/refactor it a bit.

Everything should be tested for correct functionality as well as coverage.

**Review by rpreen — APPROVED** [review]

File: sacroml/attacks/structural_attack.py [hunk 1]
```diff
@@ -71,15 +87,15 @@ class StructuralAttackResults:
...
+def get_unnecessary_risk(model: BaseEstimator | torch.nn.Module) -> bool:
     """Check whether model hyperparameters are in the top 20% most risky.
 
      This check is based on a classifier trained on results from a large
      scale study described in: https://doi.org/10.48550/arXiv.2502.09396
 
     Parameters
     ----------
-    model : BaseEstimator
+    model : BaseEstimator|pytorch.nn.Module
```
rpreen: Should this be `torch.nn.Module`?

jim-smith: @rpreen yes, will change

PR #421, opened 2026-03-06
fix: replace NaN and Infinity with null in JSON reports

### LINKED ISSUE #366: [BUG]invalid characters in json reports produced by attacks
**Describe the bug**
Running the attacks crates a json file which in some circumstances has tokens which are not part of the basic json standard and so won;t be parsed by some browsers
- NaN instead of Null
- Infinity.

**To Reproduce**
1. go to the folder examples/sklearn/cancer
2. python -m train_rf_cancer
3.python -m attack_rf_cancer
4. this produces output folder, but if you try ot open report.json you get "JSON Parse error: Unexpected identifier "Infinity""

**Expected behavior**
json report should not contain invalid tokens.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: linux (GEL) or Mac (home)
 - Browser firefox (GEL) or safari (home)
 - Version : sacroml v1.4

**Additional context**
1:
xgboost parameters sometimes come across as NaN, 
Internet suggest easiest fix is to edit ```sacroml.attacks.report.write_json()```  so that after creating the json output as a string, we do a replace() on contents of that string replacing NaN with null

2. Sometimes get Infinity as first line in "roc_thresh" - not sure where it is most appropriate to check and replace that value - possible in metrics.py

ssrhaso:
Closes #366

JSON reports produced by attacks could contain bare `NaN`, `Infinity`, 
and `-Infinity` tokens which are invalid per the JSON spec and cause 
parse errors in browsers.

Changes:
- Added float nan/inf guard in `CustomJSONEncoder.default()`
- Sanitised output string in `write_json()` via `.replace()`
- Added test to verify non-finite floats serialise as `null`

### DIFF
--- a/sacroml/attacks/report.py
+++ b/sacroml/attacks/report.py
@@ -126,6 +126,9 @@ def write_json(output: dict, dest: str) -> None:
     """Write attack report to JSON."""
     attack_formatter = GenerateJSONModule(dest + ".json")
     attack_report: str = json.dumps(output, cls=CustomJSONEncoder)
+    attack_report = attack_report.replace("-Infinity", "null")
+    attack_report = attack_report.replace("Infinity", "null")
+    attack_report = attack_report.replace("NaN", "null")
     attack_name: str = output["metadata"]["attack_name"]
     attack_formatter.add_attack_output(attack_report, attack_name)
 
--- a/sacroml/attacks/report.py
+++ b/sacroml/attacks/report.py
@@ -122,13 +122,32 @@ STRUCTURAL_GLOSSARY = {
 }
 
 
+def _sanitise_floats(obj: Any) -> Any:
+    """Recursively replace non-finite floats with None.
+
+    Parameters
+    ----------
+    obj : Any
+        Object to sanitise.
+
+    Returns
+    -------
+    Any
+        Sanitised object with non-finite floats replaced by None.
+    """
+    if isinstance(obj, float) and (np.isnan(obj) or np.isinf(obj)):
+        return None
+    if isinstance(obj, dict):
+        return {k: _sanitise_floats(v) for k, v in obj.items()}
+    if isinstance(obj, list):
+        return [_sanitise_floats(v) for v in obj]
+    return obj
+
+
 def write_json(output: dict, dest: str) -> None:
     """Write attack report to JSON."""
     attack_formatter = GenerateJSONModule(dest + ".json")
-    attack_report: str = json.dumps(output, cls=CustomJSONEncoder)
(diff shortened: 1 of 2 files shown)


sacroml/attacks/report.py
```diff
@@ -126,6 +126,9 @@ def write_json(output: dict, dest: str) -> None:
     """Write attack report to JSON."""
     attack_formatter = GenerateJSONModule(dest + ".json")
     attack_report: str = json.dumps(output, cls=CustomJSONEncoder)
+    attack_report = attack_report.replace("-Infinity", "null")
+    attack_report = attack_report.replace("Infinity", "null")
+    attack_report = attack_report.replace("NaN", "null")
```
rpreen: I believe that:
1. this wont catch the case where it's "nan" instead of "NaN"
2. if any part of the metrics include this text it will replace them too; try adding `"BaNaNa": "BaNaNa"` to the test data for example
3. it seems safer to handle this problem in the JSON encoder?

rpreen:
> * Added float nan/inf guard in `CustomJSONEncoder.default()`

The code does not match your comments.

ssrhaso:
> > * Added float nan/inf guard in `CustomJSONEncoder.default()`
> 
> The code does not match your comments.

Apologies for any errors I have made on this issue, I have tried to implement your suggested changes, please let me know your thoughts @rpreen

rpreen:
I don't think this will work for numpy arrays.

Try adding the test case:
```python
"array": np.array([1.0, np.nan, np.inf])
```

which should assert to the following:
```python
assert inner["array"] == [1.0, None, None]
```

ssrhaso:
> I don't think this will work for numpy arrays.
> 
> Try adding the test case:
> 
> ```python
> "array": np.array([1.0, np.nan, np.inf])
> ```
> 
> which should assert to the following:
> 
> ```python
> assert inner["array"] == [1.0, None, None]
> ```

Changes made and updated branch. I couldn't think of a different way other than the one you suggested for numpy array handling but lgtm. Let me know your thoughts @rpreen

jim-smith (commented):
This looks reasonable to me. 
@rpreen ?
For reference, I encountered this problem when looking at results.json produced from running attacks on some very vulnerable xgboost models. 
it was always the first entries in the roc_thresholds entries - which probably arise  from a divide by zero issue

rpreen (approved):
lgtm; thanks for making the changes

rpreen:
You need to add your details to the `CITATION.cff` but you can do that in another PR - I'll just merge this one now.

PR #422, opened 2026-03-10
chore: enforce type hints across attacks, config, and safemodel modules

### LINKED ISSUE #415: [Chore] Enforce type hints
We should require type hints to specified, it improves documentation, code completion tools, and can identify unintended behaviour.

Uncomment the Ruff check.
https://github.com/AI-SDC/SACRO-ML/blob/bcc2884164cc7412052bb7bcd7d8af70e100fa8d/pyproject.toml#L98

Add any missing types (and if possible tighten existing ones).

Tests and examples can be probably be omitted (probably ANN401 will need to be disabled since we will have some `Any` types); this can be done for example in the `pyproject.toml`:
https://github.com/AI-SDC/SACRO-ML/blob/bcc2884164cc7412052bb7bcd7d8af70e100fa8d/pyproject.toml#L167-L171

shamykyzer:
**Linting & CI:**
- Enabled the Ruff `ANN` rule set to require type hints on all function signatures
- Disabled `ANN401` to permit `Any` in cases where interfaces are intentionally dynamic
- Added a `mypy` (v1.19.1) pre-commit hook for static type analysis

**Annotations:**
- Added type annotations to function signatures across 17 files in `attacks/`, `config/`, and `safemodel/`
- Refined local variable annotations — kept where they add clarity, removed where redundant

**Cleanup:**
- Removed unused imports (`PyTorchDataHandler`, `SklearnDataHandler`) from `target.py`

**Closes #415**


rpreen:
Looks like a good start - anything and everything we can lock down the better - adding them to the variable definitions inside functions too would be good because it's such an easy way to catch any unexpected behaviour.

shamykyzer:
hello @rpreen, can you have a look at this and let me know if there is anything you want me to add or change?

sacroml/attacks/utils.py
```diff
@@ -151,7 +152,7 @@ def get_shadow_model(shadow_path: str, idx: int) -> tuple[Any, np.ndarray, np.nd
 
 def get_n_shadow_models(shadow_path: str) -> int:
     """Return the number shadow models saved."""
-    count: int = 0
+    count = 0
```
rpreen: This PR seems to remove some of the existing type hints like this one that were fine.

Please review the whole diff and check carefully for the ones that have been removed.

shamykyzer: I was worried that I was adding too much irrelevant type hints, I didn't know exactly the scope that I was working so I adjusted accordingly to see what was good to include and what to ignore.

Would you like me to include more if that's the case?

rpreen:
This also needs a slight change to `report.py` to deal with the conflict from the PR just merged.

sacroml/attacks/utils.py
rpreen: I would like the type hints put back where there was nothing wrong with them and they don't cause any issues with mypy - if some fail mypy (now that's enabled) and for some reason turn out to be a complicated pain then that's maybe a reason to remove them, but ones like this int and a bunch of strings that i quickly glanced at there's no way there is anything wrong with them.

rpreen: > Would you like me to include more if that's the case?

If there are others that you saw that you can add then that would be good - there is a reason that TypeScript is dominating JavaScript on github now - type errors are a serious pain, especially in the world of LLMs. When you are adding them in and trying to set a tight scope keep an eye out if there are actual errors due to the types. In C/C++ it's not uncommon to use [const correctness](https://en.wikipedia.org/wiki/Const_(computer_programming)) to bolt down as many parts as possible - anywhere the static analysis can prevent an error is a good thing; we could do that here with `Final` but I don't think I've ever really seen a Python project do that...

shamykyzer: This makes a lot of sense, I'll commit the changes back soon

shamykyzer:
Hi @rpreen , I was going through the type hints and ran into a couple of things worth noting:

- Renamed `indices_train` to `train_set` in `likelihood_attack.py:248` — mypy flagged it as a redefinition because `indices_train` is first returned as `np.ndarray` from `get_shadow_model` on line 239, then converted to `set[int]` on line 248. Renaming avoids reusing the same name with two different types.
- Wrapped `min_child_weight` in `structural_attack.py:206` to fit the 88-char line limit in ruff.

Also found what I think is a pre-existing bug: `get_shadow_model` in `utils.py:148` loads `indices_train.pkl` twice, the second one should be `indices_test.pkl`.

It works correctly when saving but not during loading, was this intentional?

rpreen:
> Also found what I think is a pre-existing bug: `get_shadow_model` in `utils.py:148` loads `indices_train.pkl` twice, the second one should be `indices_test.pkl`.

Good find - we need a bug fix PR for this as high priority.

rpreen:
> > Also found what I think is a pre-existing bug: `get_shadow_model` in `utils.py:148` loads `indices_train.pkl` twice, the second one should be `indices_test.pkl`.
> 
> Good find - we need a bug fix PR for this as high priority.

Thankfully it looks like this doesn't actually effect anything because the test indices aren't used when computing the LiRA scores (the check is simply if not in train indices) - but still important to fix in case they get used in future code.

jim-smith:
@shamykyzer nice find!.   this is why collaborative s/w is best

shamykyzer:
Hello @rpreen, can you review this PR when you get a chance?

  - Type hints are at 99% coverage with 0 missing annotation violations from ruff and 16 remaining are ANN401 in 4 files for dynamic interfaces.
                                                                                                                                                     
Should I expand the mypy files list beyond the current 8 files and are the 16 Any usages fine or should I narrow them?

Also, current ruff configuration only flags missing hints, wouldn't it be worth adding monkeytype or autotyping to auto-generate them going forward?
                                                                                                                                                     
  Or is it good as it is? Thanks.

rpreen:
This looks good as it is I think - I don't think we need to have everything perfectly typed, just adding as much as we can sensibly helps. I don't know enough about auto-generating types to comment on whether that would be useful to add - but I would think that is not something that would be added to Ruff/pre-commit, but would be done in a special PR like this (or a future one) and carefully reviewed. I suspect that since you have 99% done here that an auto-generator would likely struggle with the remaining complex cases and possibly just put `Any` for the areas we currently do need dynamic types anyway.

My only last question here: is there a reason for removing the existing types from `sacroml/config/target.py` and `sacroml/config/attack.py`?

We can expand more files later.

shamykyzer:
> This looks good as it is I think - I don't think we need to have everything perfectly typed, just adding as much as we can sensibly helps. I don't know enough about auto-generating types to comment on whether that would be useful to add - but I would think that is not something that would be added to Ruff/pre-commit, but would be done in a special PR like this (or a future one) and carefully reviewed. I suspect that since you have 99% done here that an auto-generator would likely struggle with the remaining complex cases and possibly just put `Any` for the areas we currently do need dynamic types anyway.
> 
> My only last question here: is there a reason for removing the existing types from `sacroml/config/target.py` and `sacroml/config/attack.py`?
> 
> We can expand more files later.

                                                                                                                                      
  Well, I thought since every function now has type annotations on its parameters and return values, mypy will infer the local variable types from the function call so annotating the variable too would've been redundant.                                                                                         
                  
```
  def _get_defaults(name: str) -> dict:                                                                                                                    
      ...                                                                                                                                                  
   
  params: dict = _get_defaults(name)  # redundant — mypy already knows it's a dict                                                                         
  params = _get_defaults(name)        # mypy infers dict from the return type
```
                                                                                                                                                           
  Here `_get_defaults` is annotated with `-> dict`, so when `mypy` sees `params = _get_defaults(name)` it already knows params is a `dict` from the return type and adding `: dict` on the variable just states the same thing twice. 
  
  Same with `name: str = prompt(...)`
  the `prompt()` returns `str`, so mypy will already know name is a `str` without the explicit annotation.
                                                       
  This applies to all the removed annotations in both files, happy to restore them if you prefer?
  
  Also I agree on expanding later, I think the `safemodel/classifiers` would be a good candidate for a future issue.

sacroml/config/attack.py
```diff
@@ -35,7 +35,7 @@ def _prompt_for_params(params: dict) -> None:
 
 def _get_attack(name: str) -> dict:
     """Get an attack configuration."""
-    params: dict = _get_defaults(name)
```
jim-smith, line 38: don't think I agree with removing type hints here in fact I think we should probably be enforcing what type of dict it is

sacroml/config/attack.py
```diff
@@ -49,7 +49,7 @@ def _prompt_for_attacks() -> list[dict]:
     while utils.get_bool("Would you like to add an attack?"):
         while True:
             print(f"Attacks available: {', '.join(names)}")
-            name: str = prompt("Which attack?: ", completer=completer)
+            name = prompt("Which attack?: ", completer=completer)
```
jim-smith: why remove the type hint here?

sacroml/config/attack.py
```diff
@@ -64,8 +64,8 @@ def _default_config() -> list[dict]:
     attacks: list[dict] = []
     names: list[str] = list(factory.registry.keys())
     for name in names:
-        params: dict = _get_defaults(name)
-        attack: dict = {"name": name, "params": params}
+        params = _get_defaults(name)
+        attack = {"name": name, "params": params}
```
jim-smith: again why remove the type hints? do they do harm?

sacroml/config/attack.py
```diff
@@ -82,7 +82,7 @@ def prompt_for_attack() -> None:
         attacks = _prompt_for_attacks()
 
     # write to file
-    filename: str = "attack.yaml"
+    filename = "attack.yaml"
```
jim-smith: why remove type hints? don't they guard against future incorrect assignment?

sacroml/config/target.py
```diff
@@ -165,7 +165,7 @@ def _get_params() -> dict[str, Any]:
     print("Enter hyperparameter names and values.")
     print("Type 'done' as the name when you're finished.")
     while True:
-        name: str = prompt("Hyperparameter name (or 'done' to finish): ").strip()
+        name = prompt("Hyperparameter name (or 'done' to finish): ").strip()
```
jim-smith: why remove type hints? don't they guard against future incorrect assignment?

sacroml/config/target.py
```diff
@@ -285,19 +285,19 @@ def prompt_for_target() -> None:
         sys.exit()
 
     # Check target directory exists and create as necessary
-    target_dir: str = "target"
-    path: str = utils.check_dir(target_dir)
+    target_dir = "target"
+    path = utils.check_dir(target_dir)
```
jim-smith: why remove type hints? don't they guard against future incorrect assignment?

sacroml/config/target.py
```diff
@@ -285,19 +285,19 @@ def prompt_for_target() -> None:
...
 
     # Get model and training information
-    model_type: str = _get_model_type()
-    model_path: str = _get_model_path()
+    model_type = _get_model_type()
```
jim-smith: why remove type hints? don't they guard against future incorrect assignment?

sacroml/config/target.py
```diff
@@ -285,19 +285,19 @@ def prompt_for_target() -> None:
...
+    model_path = _get_model_path()
 
     if model_type == "PytorchModel":
-        model_module_path: str = _get_model_module_path()
-        model_name: str = _get_model_name(path=model_module_path)
-        model_params: dict = _get_model_params()
-        train_module_path: str = _get_train_module_path()
-        train_params: dict = _get_train_params()
+        model_module_path = _get_model_module_path()
```
jim-smith: why remove type hints? don't they guard against future incorrect assignment?

sacroml/safemodel/safemodel.py
```diff
@@ -282,7 +282,7 @@ def save(self, name: str = "undefined") -> None:
 
     def __get_constraints(self) -> dict:
         """Get constraints relevant to the model type from the a read-only file."""
-        rules: dict = {}
+        rules = {}
```
jim-smith: why remove type hints? don't they guard against future incorrect assignment?

sacroml/safemodel/safemodel.py
```diff
@@ -594,7 +594,7 @@ def request_release(self, path: str, ext: str, target: Target = None) -> None:
         msg_prel, disclosive_prel = self.preliminary_check(verbose=False)
         msg_post, disclosive_post = self.posthoc_check()
         # prepare results
-        output: dict = {
+        output = {
```
jim-smith: why remove type hints? in fact couldn't we be more specific about what is supposed be in the dicts?
I know python is dynamic typing but having the hints in helps coders guard against future incorrect assignment?

jim-smith (changes requested):

jim-smith:
@shamykyzer please will  you raise issues for the glitches you spotted in ther lira code,

shamykyzer:
Hi @jim-smith, understood and fixed all the areas you highlighted across commits e1b29c7, e243639, and c929e05 restoring the annotations and the two flagged dicts tightened to ```dict[str, Any]```
                
Also raised #442 for the `get_shadow_model` double read of `indices_train.pkl`

jim-smith:
> Hi @jim-smith, understood and fixed all the areas you highlighted across commits [e1b29c7](https://github.com/AI-SDC/SACRO-ML/commit/e1b29c7f0e28710397ffc9e7563366e26d639af9), [e243639](https://github.com/AI-SDC/SACRO-ML/commit/e24363927d07abe9275c977735a41e135a9b4eca), and [c929e05](https://github.com/AI-SDC/SACRO-ML/commit/c929e052ce427394538002735c8629ab281c977d) restoring the annotations and the two flagged dicts tightened to `dict[str, Any]`
> 
> Also raised #442 for the `get_shadow_model` double read of `indices_train.pkl`

thanks @shamykyzer much appreciated

sacroml/attacks/target.py
```diff
@@ -62,7 +66,8 @@ class Target:
 
     Attributes
     ----------
-    model : Any
+    model : SklearnModel | PytorchModel | sklearn.base.BaseEstimator |
+        torch.nn.Module | None
```
jim-smith: could this be `model:ModelLike` so that if we want to add support for tensorflow etc we only have to make changes in one place?

jim-smith (approved):
@shamykyzer just  one last edit to use `ModeLike` in one place in target.py please.   But I've approved so you should be able ot merge after that.  thank you

jim-smith (approved):

PR #424, opened 2026-03-11
fix: Replace network calls to fetch data from OpenML with local data 

shamykyzer:
### Summary
- Replaced all `fetch_openml` network calls with a deterministic local `_make_local_nursery_data()` helper using `sklearn.datasets.make_classification`, eliminating flaky CI failures caused by OpenML server unavailability
- Added `_make_local_nursery_data()` in `tests/conftest.py`, `examples/sklearn/nursery/dataset.py`, and `examples/safemodel/safemodel.py` to generate synthetic nursery-like categorical data locally
- Updated `tests/attacks/sklearn_dataset.py` to import the shared helper from `conftest.py` instead of calling OpenML
- Relaxed metric assertions in `tests/attacks/test_factory.py` to use range-based checks (model score >= 0.8, TPR >= 0.9, 0.3 <= FPR <= 0.5) since synthetic data has different separability than the original dataset
- Added `tests/test_openml_dependency.py` guard test that scans the test directory to prevent reintroduction of `fetch_openml`
- Cleaned up stale OpenML references in docstrings across examples


shamykyzer:
### Summary
- Replaced all `fetch_openml` network calls with a deterministic local `_make_local_nursery_data()` helper using `sklearn.datasets.make_classification`, eliminating flaky CI failures caused by OpenML server unavailability
- Added `_make_local_nursery_data()` in `tests/conftest.py`, `examples/sklearn/nursery/dataset.py`, and `examples/safemodel/safemodel.py` to generate synthetic nursery-like categorical data locally
- Updated `tests/attacks/sklearn_dataset.py` to import the shared helper from `conftest.py` instead of calling OpenML
- Relaxed metric assertions in `tests/attacks/test_factory.py` to use range-based checks (model score >= 0.8, TPR >= 0.9, 0.3 <= FPR <= 0.5) since synthetic data has different separability than the original dataset
- Added `tests/test_openml_dependency.py` guard test that scans the test directory to prevent reintroduction of `fetch_openml`
- Cleaned up stale OpenML references in docstrings across examples

shamykyzer:
Hello @jim-smith, sorry for the confusion about the other PR.

This is the PR that replaces all fetch_openml() calls with locally generated synthetic data using sklearn.datasets.make_classification(), so tests no longer depend on network access to OpenML.

There's also a guard test (test_openml_dependency.py) that will fail if fetch_openml is reintroduced in any test file. 

Could you review the changes and let me know your thoughts?

examples/sklearn/nursery/dataset.py
jim-smith, line 1: Code looks fine.
Thinking about ongoing maintenance, it would be cleaner to pull the code for `_make_local_nursery_data()` into a separate file within examples rather than duplicate it in two files in this folder?

tests/attacks/test_factory.py
```diff
@@ -49,5 +50,6 @@ def test_factory(monkeypatch, get_target):
     metrics = report[nr]["attack_experiment_logger"]["attack_instance_logger"][
         "instance_0"
     ]
-    assert metrics["TPR"] == pytest.approx(0.91, abs=0.01)
-    assert metrics["FPR"] == pytest.approx(0.41, abs=0.01)
```
jim-smith, line 53: is it possible to be more precise about the expected values ?
-  (e.g. by fixing the `random_state` input value when the fake nursery data is generated)

jim-smith: it gives more confidence the other code (LiRA in this case) is w working as expected

jim-smith (changes requested):

shamykyzer:
## Summary
- Created `examples/nursery_data.py` with the shared `_make_local_nursery_data()` function
- Updated `examples/safemodel/safemodel.py` to import from it instead of duplicating around 50 lines
- `examples/sklearn/nursery/dataset.py` keeps its own copy because `Target.save()` copies this file standalone, it I think it must be left self-contained

Also changed the dummy feature generation in `get_target` to use a local `np.random.RandomState(1)` instead of the global RNG, so the values are always the same regardless of test order.

I ran the full test pipeline twice with all random state pinned (`np.random.seed(1)`, `random_state=1` in data generation, splits, and the RF model) both runs gave identical results (TPR=0.954, FPR=0.463) so
- TPR assertion now uses `pytest.approx(0.95, abs=0.01)`
- FPR assertion now uses `pytest.approx(0.46, abs=0.01)`

jim-smith (commented):
Good to have fixed the random state.

Still have a nagging doubt about having the function that creates the data replicated in three different places , but maybe that is not problematic.

rpreen:
I don't think the examples should be modified here - they are meant to be very simple and this adds a lot of data processing code, it also serves as a nice example for how OpenML datasets can be used. 

This issue was really about solving CI testing which can have network issues.

rpreen:
I don't see the point in have a test entirely dedicated to checking OpenML is not used.

PR #425, opened 2026-03-11
fix: prevent data arrays being serialized on save when dataset module is provided

### LINKED ISSUE #411: [Task] Review data handling
Currently data for PyTorch models is converted into numpy and stored in RAM due to legacy reasons that the attacks were originally specifically designed for scikit-learn; there may be a better way. It could do with a review to see if it can be improved.

Also, when providing a dataset module to the `Target` class and calling `save()`, it will save just the code (correctly) but when the `Target` is subsequently loaded from storage, the data is converted to numpy to populate variables such as `X_train` (as mentioned above) --- this means that a subsequent call to `save()` will write the full dataset as pkl, which is not intended. I think this should be a simple fix by adding a check to `target.py`.

ssrhaso:
Fixes #411 (Part B)

When a `Target` is loaded from storage with a `dataset_module_path` set, `load()` calls `load_sklearn_dataset()` or `load_pytorch_dataset()` which populates `X_train`, `y_train`, `X_test`, `y_test` in memory. A subsequent call to `save()` then finds those arrays non-null and serializes them as `.pkl` files via `_save_array()` - defeating the purpose of using a dataset module.

## Fix
Added a guard in `_save_array()` that skips serialization of dataset-backed arrays (those defined in the `ARRAYS` constant) when `dataset_module_path` is set. Non-dataset attributes (`proba_train`, `proba_test`, `indices_train`, `indices_test`) are unaffected.

## Changes
- `sacroml/attacks/target.py` - guard added to `_save_array()`
- `tests/attacks/test_attacks_target.py` - regression test added

## Notes
Part A of #411 (PyTorch in-RAM numpy conversion) is a larger architectural refactor and is out of scope for this PR.

I believe to solve the former part of this issue it would need a redesign to  the internal data contract so attacks can accept DataLoaders (or lazy tensors) in addition to numpy arrays. I believe this would require refactoring ~100 lines of code in attacks/ , utils.py

@rpreen

### DIFF
--- a/sacroml/attacks/target.py
+++ b/sacroml/attacks/target.py
@@ -390,6 +390,13 @@ class Target:
 
     def _save_array(self, path: str, target: dict, attr_name: str) -> None:
         """Save numpy array as pickle."""
+        # When a dataset module is provided the data arrays are re-populated
+        # from it on every load(); serialising them would defeat the purpose
+        # of using a module and cause a re-save to write the full dataset to
+        # disk
+        if self.dataset_module_path and attr_name in ARRAYS:
+            target[f"{attr_name}_path"] = ""
+            return
         arr = getattr(self, attr_name)
         if arr is not None:
             arr_path = os.path.join(path, f"{attr_name}.pkl")
--- a/tests/attacks/test_attacks_target.py
+++ b/tests/attacks/test_attacks_target.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+import os
+
 import numpy as np
 import pytest
 from sklearn.ensemble import RandomForestClassifier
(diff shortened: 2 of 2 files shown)


ssrhaso:
@rpreen Tackled only part B of the initial issue as I believe the former should be considered for a broader issue scope - much refactoring of many files may be necessary as far as I can tell. Please correct me if this is an incorrect way to approach this issue and if there is any way to directly improve upon this.

sacroml/attacks/target.py
```diff
@@ -390,6 +390,13 @@ def _save_dataset_module(self, path: str, target: dict) -> None:
 
     def _save_array(self, path: str, target: dict, attr_name: str) -> None:
         """Save numpy array as pickle."""
+        # When a dataset module is provided the data arrays are re-populated
+        # from it on every load(); serialising them would defeat the purpose
+        # of using a module and cause a re-save to write the full dataset to
+        # disk (issue #411).
```
rpreen: Please remove the reference to the issue number - if every issue was referenced in the code it would add a lot of junk - same goes for the test docstring.

rpreen (approved):
lgtm

PR #429, opened 2026-03-24
test: clean up test warnings and resolve matplotlib backend issue

### LINKED ISSUE #404: [Tests] Clean up tests and review outstanding warnings
some of the warnings about future changes in numpy and new starting to manifest as exceptions being raised.
Needs someone to go through and review the warnings and adapt as necessary

ssrhaso:
Closes #404

Ran full test suite with `-W all` on Python 3.12 / numpy 2.4.2 / sklearn 1.8.0.

No numpy deprecation or future warnings found,  those may have been resolved by dependency updates since the issue was opened. All warnings present were RuntimeWarnings.

Before: 89 warnings, 2 test-ordering failures. After: 0 warnings, 136 passed.

The following are changes I made:

- `tests/conftest.py`: force matplotlib Agg backend to prevent Tk failures during full suite runs
- `sacroml/attacks/utils.py`: guard `shapiro()` with `len >= 8` and suppress degenerate-sample warnings
- `sacroml/attacks/report.py`: close figures after saving (×2)
- `sacroml/attacks/attribute_attack.py`: close figures after saving (×2)
- `tests/attacks/test_sklearn.py`: ravel column vector before `LabelEncoder`
- `tests/attacks/test_structural_attack.py`: filter `ConvergenceWarning` on MLP tests (×2)

- added myself to citation as per prior PR merged as advised by rpreen.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,8 @@
-"""Common utility functions for testing."""
+"""Pytest configuration for the test suite."""
+
+import matplotlib as mpl
+
+mpl.use("Agg")
 
 import contextlib
 import os
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -21,6 +21,14 @@ from sacroml.attacks.target import Target
 
 np.random.seed(1)
 
+
+def pytest_sessionfinish(_session, _exitstatus):
+    """Remove test-created target directories after all tests complete."""
+    for folder in ("target_pytorch", "target_sklearn"):
+        with contextlib.suppress(Exception):
+            shutil.rmtree(folder)
+
+
 folders = [
     "RES",
     "dt.sav",
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -22,7 +22,7 @@ from sacroml.attacks.target import Target
 np.random.seed(1)
 
 
-def pytest_sessionfinish(_session, _exitstatus):
+def pytest_sessionfinish(session, exitstatus):  # noqa: ARG001
     """Remove test-created target directories after all tests complete."""
     for folder in ("target_pytorch", "target_sklearn"):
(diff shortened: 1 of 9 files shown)


rpreen:
These changes look good.

However, when running code coverage there are a couple of `CoverageWarning` messages that could also be cleaned up here (see the [CI run](https://github.com/AI-SDC/SACRO-ML/actions/runs/23496933578/job/68380581417?pr=429)) - I think the folders `target_pytorch` and `target_sklearn` just need to be deleted after the test runs to prevent them from being included in the coverage report.

You can generate the coverage locally if you wish to test:
```shell
pytest --cov=./
```

Also, there are still 16 warnings reported on the CI - although it's only printing a message for one of them. For some reason I get 88 warnings running locally with the same Python version as the CI., but since they will vary across versions/dependencies perhaps we should stick to the CI as the main concern.

ssrhaso:
I Addressed both points, scoped CI coverage to --cov=sacroml --cov=tests to eliminate the CoverageWarnings, and added a filterwarnings entry for the multiprocess fork DeprecationWarning. 

CI now shows 0 warnings, let me know @rpreen

tests/conftest.py
```diff
@@ -17,6 +21,14 @@
 
 np.random.seed(1)
 
+
+def pytest_sessionfinish(session, exitstatus):  # noqa: ARG001
```
rpreen: Can you move the rule suppression into `pyproject.toml` per-file-ignores? Inline pragmas clutter things and it's easier to keep track of them all if they are in a central location.

ssrhaso: Done, apologies I am still becoming familiar with the repository and structure

rpreen (approved):
thanks

PR #430, opened 2026-03-26
test: replace fetch_openml() with local data make_classification()

### LINKED ISSUE #410: [Tests] Replace network calls to fetch data from OpenML with local data
Frequently the CI fails running tests due to network issues fetching data from OpenML, it would be good to replace calls to `fetch_openml()` with some local data, ideally created with sklearn's `make_classification()`.

shamykyzer:
- Replaces `fetch_openml()` calls in test fixtures with locally generated data using `sklearn.make_classification()` to avoid CI failures from       
  network issues.                                                                                                                                    
                                                                                                                                                     
  Closes #410


shamykyzer:
Hello @rpreen, could you please have a look at this?                                                                                    
                                                                                                                                                   
  1. **test_factory.py** - The TPR/FPR/score assertions are hardcoded to the real nursery dataset values. With synthetic data the metrics shift (TPR 0.958 vs expected 0.91).
  
  Can I widen the tolerances or use range checks since this test verifies the factory pipeline, not model performance?
                                                                                                                                                     
  2. **test_adaboost_nondisclosive** - `min_samples_leaf=200` with ~40 training samples means the tree can never split. But AdaBoost rejects it as worse than random. 
  
  Also I believe the adaboost test is a pre-existing bug: `min_samples_leaf=200` on ~40 samples means the tree can never split, and the missing `random_state` makes it pass or fail depending on test ordering.
  
  Can you confirm that I am not misunderstanding?
  
  Thanks

rpreen:
If there's a random seed set, does the test number need to be widened or just slightly adjusted? If a slight tweak is needed because it's using synth data that seems fine to me.

The AdaBoost test should also have the seed set (they all should really). It looks like originally the disclosive and non-disclosive tests were in a single function which did set the seed to 42 (in fact I explicitly added the setting of the seed in #318 because it was causing problems) but then in #367 it was split into two functions and the random seed was removed and there were some other parameter changes which may have changed the behaviour. Perhaps you can use the original as a guide since - @jim-smith will know more than me as I didn't work on that test.

rpreen:
There must be some way of avoiding the code duplication with the data creation?

rpreen (approved):
lgtm; thanks for the changes

PR #431, opened 2026-03-27
feat: new instance-based attack for data leakage in SVM/kNN models

### LINKED ISSUE #454: chore: define shared atol semantics across attacks
Jim asked on #431 [issue](https://github.com/AI-SDC/SACRO-ML/pull/431#discussion_r3138161139) for the `atol=1e-8` in `InstanceBasedAttack` to be globally defined, alongside whatever `StructuralAttack` uses for equivalence classes

`InstanceBasedAttack` uses `np.allclose(atol=1e-8)`, `StructuralAttack` uses `np.unique(axis=0)` exact equality, so before extracting a constant I wanted to check whether you would like one shared tolerance across attacks, or two distinct constants with different semantics

@rpreen what do you think

shamykyzer:
- New `InstanceBasedAttack` detects training data leakage in models that store raw instances, support vectors (`SVC`, `NuSVC`, `OneClassSVM`) and stored neighbours (`KNeighborsClassifier`, `KNeighborsRegressor`)
- Compares stored instances against training data via `np.allclose`, reports first ten matches with feature previews, plus storage fraction and match fraction metrics
- Unwraps `sklearn.Pipeline` so the comparison runs in the final estimator's feature space
- Detects differentially private model variants and surfaces a mitigation note
- Registered in the factory under `"instance_based"`
- Match Fraction glossary updated to "A non-zero match fraction confirms data leakage" per Jim's review
- Module level constants extracted: `INSTANCE_MATCH_ATOL = 1e-8`, `N_EXAMPLES = 10`, `N_FEATURE_PREVIEW = 10`

Closes [[New Feature Request] New Attack: Model contains training data#59](https://github.com/AI-SDC/TREvolution_disclosure_control_work/issues/59)
Closes #454


sacroml/attacks/instance_based_attack.py
```diff
@@ -0,0 +1,542 @@
...
+        "noise to break the direct link between stored model parameters "
+        "and the original training data, mitigating the leakage risk."
+    ),
+    "Storage Fraction": (
+        "The proportion of training data points stored inside the model. "
+        "For SVM this is typically a subset; for kNN this is 1.0 (all data)."
+    ),
+    "Match Fraction": (
+        "The proportion of stored instances that exactly match a training "
+        "data point. A high match fraction confirms data leakage."
```
jim-smith: 'A high match fraction ...' => 'A non-zero match fraction...'

becuase even one data point leaked is to many :)

sacroml/attacks/instance_based_attack.py
```diff
@@ -0,0 +1,542 @@
...
+    extracts those stored instances, compares them to the training data, and
+    reports whether the model leaks training data.
+    """
+
+    def __init__(
+        self,
+        output_dir: str = "outputs",
+        write_report: bool = True,
+        n_examples: int = 10,
+        atol: float = 1e-8,
```
jim-smith: never a fan of seeing numbers in code :) - but this raises an important issue that applies to structural attacks we calculate equivalent classes.
The default value of `atom`  should probably be globally defined. 
Might need to discuss with @rpreen and raise this as a separate issue

sacroml/attacks/instance_based_attack.py
```diff
@@ -0,0 +1,542 @@
...
+        if not target.has_model():
+            logger.info("target.model is missing, cannot proceed")
+            return False
+        if not target.has_data():
+            logger.info("target data is missing, cannot proceed")
+            return False
+        return True
+
+    @staticmethod
+    def _unwrap_model(model):
```
jim-smith: Think this function is really useful and should perhaps be incorporated into other attacks. 
As a starting point for reusability, perhaps it should be moved to utils.py?

jim-smith (changes requested):
@shamykyzer Just a couple of minor changes to make please.
If you want to raise the 'move dealing with pipelines into utils.py' as a separate issue, and leave that code in here for now that is fine.

shamykyzer:
Hi @jim-smith, could you please review this PR so far? 

thanks a lot.

sacroml/attacks/instance_based_attack.py
```diff
@@ -0,0 +1,560 @@
...
+
+        Parameters
+        ----------
+        output_dir : str
+            Name of a directory to write outputs.
+        write_report : bool
+            Whether to generate a JSON and PDF report.
+        n_examples : int
+            Maximum number of matching examples to include in the report.
+        atol : float
```
jim-smith, line 151: why is there no option to report_individual as we have for most of the other attacks ?
Please can we have that

sacroml/attacks/instance_based_attack.py
```diff
@@ -0,0 +1,560 @@
...
+            Indices of stored instances into the original training data.
+        X_train : np.ndarray
+            The training data to compare against.
+
+        Returns
+        -------
+        n_matched : int
+            Number of stored instances that match training data.
+        example_matches : list[dict]
+            Details of the first n_examples matches.
```
jim-smith: probably cleaner (and will automatically support the report_individual option ) if this list is just as long as it needs to be - I.e. all the matches-  then in the code that produces the report you just show the first few.

sacroml/attacks/instance_based_attack.py
```diff
@@ -0,0 +1,560 @@
...
+        "data point. A non-zero match fraction confirms data leakage."
+    ),
+}
+
+
+@dataclass
+class InstanceBasedAttackResults:
+    """Results of an instance-based model attack."""
+
+    model_type: str
```
jim-smith, line 96: iff you put default values in here you can make the code much shorter later on

sacroml/attacks/instance_based_attack.py
```diff
@@ -0,0 +1,560 @@
...
+            X_train = preprocessor.transform(X_train)
+        n_training = len(X_train)
+
+        if not is_instance_based:
+            logger.info(
+                "Model type %s is not instance-based, no data leakage risk "
+                "from stored instances.",
+                model_type,
+            )
+            self.results = InstanceBasedAttackResults(
```
jim-smith, line 312: wouldn't need most of these lines if you had used defaults when you declared  the class ?

sacroml/attacks/instance_based_attack.py
```diff
@@ -0,0 +1,560 @@
...
+            else:
+                logger.warning(
+                    "kNN model %s does not have _fit_X attribute. "
+                    "It may not be fitted.",
+                    model_type,
+                )
+
+        if stored_instances is None:
+            self.results = InstanceBasedAttackResults(
+                model_type=model_type,
```
jim-smith, line 348: see earlier comment about avoiding lostsof these tens if you have sensible defaults

sacroml/attacks/instance_based_attack.py
```diff
@@ -0,0 +1,560 @@
...
+        )
+
+        storage_fraction = n_stored / n_training if n_training > 0 else 0.0
+        match_fraction = n_matched / n_stored if n_stored > 0 else 0.0
+        data_leakage_confirmed = n_matched > 0
+
+        mitigations = self._build_mitigations(is_svm, is_knn, is_dp_safe)
+
+        self.results = InstanceBasedAttackResults(
+            model_type=model_type,
```
jim-smith, line 394: again , might not need all these if defaults have been set

sacroml/attacks/instance_based_attack.py
```diff
@@ -0,0 +1,560 @@
...
+            is_instance_based=True,
+            is_dp_safe=is_dp_safe,
+            n_stored_instances=n_stored,
+            n_training_samples=n_training,
+            storage_fraction=storage_fraction,
+            n_matched=n_matched,
+            n_checked=n_stored,
+            match_fraction=match_fraction,
+            example_matches=example_matches,
```
jim-smith: Might be more maintainable if you make consistent with other attacks I.e. check how all of the per-record results are stored and passed around in other attacks and replicate that rather than having separate array of 'examples'

jim-smith (changes requested):

sacroml/attacks/instance_based_attack.py
```diff
@@ -0,0 +1,547 @@
...
+            Indices of stored instances into the original training data.
+        X_train : np.ndarray
+            The training data to compare against.
+
+        Returns
+        -------
+        n_matched : int
+            Number of stored instances that match training data.
+        record_level_results : InstanceBasedRecordLevelResults
+            Per-stored-instance outcomes, one entry per stored instance
```
jim-smith: This is the wrong way around- it suggest the record-level outcomes have `len(stored_instances)` but it should be `len(X_train)` I.e. it is the training data which us vulnerable.

tests/attacks/test_instance_based_attack.py
```diff
@@ -0,0 +1,487 @@
...
+
+        instance = output["attack_experiment_logger"]["attack_instance_logger"][
+            "instance_0"
+        ]
+        individual = instance["individual"]
+        for key in ("stored_index", "training_index", "matched", "stored_values"):
+            assert key in individual
+
+        n_stored = instance["n_stored_instances"]
+        assert len(individual["stored_index"]) == n_stored
```
jim-smith: same Information, but not in same format as other attacks which have one record per training record (structural attacks)  or one record for the union of (train, test) (MIA attacks) measuring whether they are predicted to be in the training set.

sacroml/attacks/instance_based_attack.py
```diff
@@ -0,0 +1,547 @@
...
+
+    Each list is indexed by stored instance (an SVM support vector or a
+    kNN neighbour) and holds one entry per stored instance, matched or
+    not. Stored as parallel lists, consistent with how other attacks
+    expose per-record results, so it can be emitted under the report's
+    ``individual`` key when ``report_individual`` is set.
+    """
+
+    stored_index: list[int]
+    training_index: list[int]  # index into X_train, or -1 if no match
```
jim-smith: this is focussing on _'are the stored values present in the training set?_, but for the report we need to know _'is this training record stored in the model?'_ which is a slightly different question.

jim-smith (changes requested):
Still needs changes. I think the way it is implemented is possibly more efficient and the overall message "does this mode lcintain training instances' is answered correctly.

However,  the way that the record level results are presented (is this stored instance present in the training set) is inconsistent with the way it is presented for other attacks (which would be 'is this training record stored in the model'.  Quick change to create a new field, 
`individual_risk:np.array = np.zeros(X_train.shape[0],dtype=int)`  and then a for loop setting to `1` (True) the index of training record you have stored in individual level results

jim-smith (approved):
Thanks.
Not entirely sure why we need a data class that just holds a single sequence of boolean values (or why this is stored as a list of ints rather than as a list or numpy array of bools).

But I guess this future-proofed us?

PR #435, opened 2026-04-01
feat: add QMIAAttack quantile regression membership inference attack

### LINKED ISSUE #306: Implement quantile regression and other non-shadow attacks
respectively
https://proceedings.neurips.cc/paper_files/paper/2023/file/01328d0767830e73a612f9073e9ff15f-Paper-Conference.pdf
code at 
https://github.com/amazon-science/quantile-mia

and
https://arxiv.org/pdf/2312.03262
code at
https://github.com/privacytrustlab/ml_privacy_meter/tree/master/research/

ssrhaso:
## Summary

Implements the quantile regression membership inference attack from [Bertran et al., NeurIPS 2023] (https://arxiv.org/abs/2307.03694) - as described in #306. 

Trains a single `HistGradientBoostingRegressor` at quantile level (1 − α) on non-member hinge scores to learn per-sample membership thresholds. A record is predicted as a member when its observed score exceeds the predicted threshold. 

No shadow models required.

## New files

- `sacroml/attacks/qmia_attack.py` - `QMIAAttack` class (18 tests, 100% coverage)
- `sacroml/attacks/utils.py` - hinge score, margin conversion, label helpers
- `sacroml/attacks/report.py` - `create_qmia_report()` with `QMIA_INTRODUCTION`
- `sacroml/attacks/factory.py` - registered as `"qmia"`
- `tests/attacks/test_qmia_attack.py` - 18 tests
- `tests/attacks/test_factory.py` - factory integration test
- `tests/attacks/test_report.py` - JSON sanitisation test
- `examples/sklearn/benchmark_qmia_*.py` - benchmark scripts
- `CHANGELOG.md`, `README.md` - updated

## Incidental fixes

- `attribute_attack.py`: close matplotlib figures after saving (resource leak)
- `target.py`: skip serialising data arrays when dataset module is provided
- `test_sklearn.py`: fix `LabelEncoder` deprecation warning
- `test_structural_attack.py`: suppress `ConvergenceWarning` in MLP tests
- `conftest.py`: set matplotlib `Agg` backend, add session cleanup
- `utils.py`: refactor `check_and_update_dataset` to use dict lookup

## Test results

- 157 passed, 0 failures
- `qmia_attack.py`: 100% line coverage


tests/attacks/test_factory.py
jim-smith, line 1: I am not happy with the greatly relaxed bounds on the various attack metrics, for example going from:
```
assert metrics["TPR"] == pytest.approx(0.91, abs=0.01)
 assert metrics["FPR"] == pytest.approx(0.41, abs=0.01)
 ```
to
```
   assert 0.5 <= metrics["TPR"] <= 1.0
    assert 0.0 <= metrics["FPR"] <= 1.0
```
no longer checks that the algorithm is behaving reproducibly and the same way as previous versions,
because we are going from +/- 1%    (which allows for minor changes in precision of floats etc on different platforms)    to +/- 25%


I know this is partly to do with using randomly created data, but by specifying the  random seeds  it should still be possible to get reproducible behaviour and keep the acceptable difference in behaviour to within +/- 1% of whatever the new value is.

File: tests/attacks/test_factory.py [hunk 1]
```diff
@@ -49,5 +84,38 @@ def test_factory(monkeypatch, get_target):
...
+
+    path = os.path.normpath(f"{output_dir}/report.json")
+    with open(path, encoding="utf-8") as fp:
+        report = json.load(fp)
+
+    nr = list(report.keys())[0]
+    metrics = report[nr]["attack_experiment_logger"]["attack_instance_logger"][
+        "instance_0"
+    ]
+    assert 0 <= metrics["TPR"] <= 1
```
jim-smith: again, this is just testing that the metrics code is within valid bounds of 0 and 1 I.e. 'the algorithm  is detecting somewhere between none and all of the training records' .

But it should be behaving **reproducibly**.  I.e. 
```
metrics["TPR"] == pytest.approx(0.XX, abs=0.01)
```
for some value of XX

jim-smith: and the same for `metrics["FPR"]`
- should be constant at some value

tests/attacks/test_factory.py
shamykyzer: Hello Jim,

Sorry for the oversight. I have tightened the QMIA test assertions to `+/- 1%` using `pytest.approx`, following the same pattern as `test_factory`. I have now replaced the loose bounds with exact expected values. I also added `TPR@FPR` threshold assertions because they test QMIA in detecting members at a controlled false positive rate. I think these are worth keeping, but let me know if you think otherwise.

I went more in-depth and just wanted to confirm my understanding, as I do not want to change anything outside of the QMIA scope.

Does `test_factory` checks correctness with tight bounds while other tests just verify the code runs and produces valid metrics? and wouldn't it have been better to add `random_state` to those fixtures and tighten them from the start?

`test_lira_attack` and `test_worst_case_attack` use loose assertions like `assert 0 <= metrics["TPR"] <= 1` because their fixtures do not set `random_state`, so the metric values differ between runs. 

Since `test_factory` runs the full end-to-end pipeline and is the only test with tight assertions because its `get_target` fixture sets `random_state=1` making the results reproducible. 

The QMIA fixtures do have `random_state` set, but I kept the loose format to stay consistent with the other individual attack tests. Should I tighten those as well, like I did in the factory test?

Please let me know if I have misunderstood anything.

Thanks!

shamykyzer:
Hi @jim-smith, over the last week I pushed a few changes to this branch while wiring QMIA into MetaAttack (#441). Here is a short summary of what I changed. Could you please revise over these?

## harden QMIA regressor, calibration drift, margin rank - 61f0027

Three things in this one commit: two guards (`C1` for regressor health, `C2` for calibration) plus a margin-rank fix. Both `C1` / `C2` labels also show up in the test file comments so I can reference them from the regression tests.

**1) C1 guard: degenerate regressor check:** Sometimes the quantile regressor collapses and predicts the same threshold for every record. Before this change, QMIA would carry on and return a meaningless AUC. Now it raises a clear error.

```python
if threshold_spread < tiny_number:
    raise RuntimeError("regressor is dead")
```

A broken attack that looks like it worked is worse than one that crashes. MetaAttack would silently aggregate the bad result.

**2) C2 guard: `calibration_ok` flag:** After the attack runs, I measure the observed public FPR and compare it to `alpha`. If they are close, `calibration_ok = True`. If they drift apart, I set the flag to `False` and log a warning.

If observed FPR drifts from `alpha`, the numbers do not mean what they are supposed to, so the user needs a clear signal.

**3) Margin rank:** Each record in QMIA gets a raw "margin" number. Higher margin = more likely a member. The old code pushed every margin through a sigmoid function, which squishes any input into the 0-to-1 range. That works fine for most records, but once a margin gets above about 37, floating-point precision can no longer tell the difference between the sigmoid of 37 and the sigmoid of 100. 

Both just round to exactly 1.0 (verifiable in any Python shell: `scipy.special.expit(37) == scipy.special.expit(100) == 1.0`). So the *most vulnerable* records (the ones with the biggest margins) all come out tied at 1.0, and we lose the ability to rank them.

<details>
<summary>Reproducer (click to expand)</summary>

```python
from scipy.special import expit
import numpy as np

# Exact saturation happens at x >= 36.74 in float64
for m in [30, 35, 36, 36.74, 37, 50, 100]:
    print(f"sigmoid({m:>6}) = {expit(m)!r}")

# Rank test: eight increasing margins collapse to only two unique scores
margins = np.array([36, 36.5, 37, 37.5, 38, 40, 50, 100])
print("unique sigmoid values:", len(np.unique(expit(margins))))
```

Output:

```
sigmoid(    30) = 0.9999999999999065
sigmoid(    35) = 0.9999999999999993
sigmoid(    36) = 0.9999999999999998
sigmoid( 36.74) = 1.0
sigmoid(    37) = 1.0
sigmoid(    50) = 1.0
sigmoid(   100) = 1.0
unique sigmoid values: 2
```

</details>

```python
# before: sigmoid, tail saturates to 1.0
probs = sigmoid(margins)
# after: raw margins, tail stays distinguishable
probs = margins_to_two_column_probs(margins)
```

**Concrete proof from a fresh run** (seed 7, `class_sep=1.0` to `1.15`):

| Scenario     | QMIA `TPR@FPR=0.01` | LiRA-10 `TPR@FPR=0.01` |
|--------------|---------------------:|-----------------------:|
| n=500, d=10  | 0.170               | 0.000                  |
| n=2000, d=20 | 0.193               | 0.000                  |
| n=5000, d=30 | 0.157               | 0.000                  |

LiRA-10 hits zero because its top tail is tied across shadow-model runs. Without the margin rank fix, QMIA would hit zero for the same reason (sigmoid-saturated tail). With raw margins, QMIA recovers a meaningful tail.

As a side benefit, MetaAttack's example script ranks the top 10 most vulnerable records, which only makes sense if those records are actually distinguishable.

---

## drop extract_true_label_probs, resolve ruff errors - 3eec3bb

Dropped dead code: the `extract_true_label_probs` helper in `utils.py` and its orphan test, both unused since `50bfe11` rewired the QMIA pipeline. Also cleared a handful of ruff lint errors that were failing CI.

---

## tighten QMIA public FPR tolerance - fb313a3

Same principle as our earlier [test_factory thread](https://github.com/AI-SDC/SACRO-ML/pull/435#discussion_r3022209854), applied to the new `test_qmia_public_fpr_tracks_alpha` test I added in `61f0027`. Tightened the deviation bound so the test fails if observed FPR drifts more than `0.05` from `alpha`.

---

## warn when QMIA skips non-sklearn target remapping - 7e245ad

`check_and_update_dataset` maps class labels to `predict_proba` column positions. For non-sklearn targets (wrapped models), it cannot do that safely. Before this change, the function silently skipped the remap and carried on. Now it logs a clear warning.

I hit this while testing MetaAttack on a wrapped target. The attack produced a plausible-looking AUC computed against the wrong true-class labels. A silent skip was dangerous.

---

## reject non-finite predict_proba in QMIA - 42b06b8

Added a finite-check on `predict_proba` output before the hinge score is computed. Raises `ValueError` if any `NaN` or `Inf` is present. Added a test that monkeypatches `predict_proba` to return `NaN` and asserts the raise.

```python
if not (np.isfinite(proba_train).all() and np.isfinite(proba_test).all()):
    raise ValueError("predict_proba returned non-finite values")
```

`NaN` propagates silently through NumPy, so one bad probability poisons the whole AUC. Catching it at the gate makes the failure loud. It also protects MetaAttack's DataFrame, which averages per-record scores across reps.

---

## enable HGBR early stopping for QMIA fits with n >= 1000 - d34b9a1

Enabled `early_stopping=True` on the regressor when the non-member set has at least 1000 rows.

```python
use_early_stopping: bool = len(test_scores) >= 1000
self.quantile_model = HistGradientBoostingRegressor(
    loss="quantile",
    quantile=1.0 - self.alpha,
    max_iter=self.max_iter,
    random_state=self.random_state,
    early_stopping=use_early_stopping,
)
```

Saves roughly 20 to 40 percent of fit time on the benchmark datasets. Below 1000 rows, HGBR's internal validation split (10 percent of the data) is too small to be reliable, so the gate avoids triggering early stopping on tiny datasets.

---

```
pytest tests/attacks/test_qmia_attack.py tests/attacks/test_factory.py -v
```

sacroml/attacks/qmia_attack.py
```diff
@@ -0,0 +1,238 @@
...
+        combined_x_with_y = np.column_stack((combined_x, combined_y))
+        combined_scores = np.hstack((train_scores, test_scores))
+        thresholds = self.quantile_model.predict(combined_x_with_y)
+
+        # HGBR silently returns a constant predictor on degenerate inputs;
+        # catch that here so the attack cannot report a plausible-but-wrong AUC.
+        threshold_spread: float = float(np.std(thresholds))
+        score_spread: float = float(np.std(test_scores))
+        if threshold_spread < max(1e-10, 1e-6 * score_spread):
+            raise RuntimeError(
```
jim-smith: Appreciate what is being done here, but am not convinced that raising an Error is the correct approach when you catch this situation.
For an output checker or a researcherunnig the code it is going to look as if the code is buggy or has just 'crashed'.
- Experience with acro users  showed people are not good at looking at stack traces and interpreting them.
- So I guess the question is, what should we do when an attack can;t run? report that the model was vulnerable, or that it wasn't?    
- My suggestion (partially informed by the folks who are doing ther 'RO-CRATES' part of the project)is that we should write into the pdf and the .json files that we tried to run the attack and whether it ran successfully or not (regardless of what the outcome of a successful run was) 
-  and for a run that did not complete successfully do not produce any of the standard attack fields in the json.
- then the meta attack should only try to extract results from attacks that completed successfuilly
- this approach should extend to other types of attacks.

tests/attacks/test_factory.py
jim-smith: Hi @shamykyzer good catch. I think awe should always set random state as you suggest

tests/attacks/test_qmia_attack.py
```diff
@@ -0,0 +1,453 @@
...
+    qmia_degenerate_target: Target, tmp_path: Path
+) -> None:
+    """C1: QMIA must raise when the quantile regressor collapses to a constant."""
+    attack_obj: QMIAAttack = QMIAAttack(
+        output_dir=str(tmp_path / "qmia_degen"),
+        write_report=False,
+        alpha=0.01,
+    )
+
+    with pytest.raises(RuntimeError, match="degenerated to a near-constant"):
```
jim-smith: might be affected by my earlier comment about whether we should be raising errors

examples/sklearn/benchmark_qmia_full.py
```diff
@@ -0,0 +1,239 @@
...
+from sacroml.attacks.likelihood_attack import LIRAAttack
+from sacroml.attacks.qmia_attack import QMIAAttack
+from sacroml.attacks.target import Target
+from sacroml.attacks.worst_case_attack import WorstCaseAttack
+
+logging.disable(logging.CRITICAL)
+warnings.filterwarnings("ignore")
+
+
+def _make_target(x, y, name):
```
jim-smith: missing function docstrings?

jim-smith (changes requested):

shamykyzer:
Hi @jim-smith, thanks again for the detailed review. I've worked through your comments. Summary below, and I'd appreciate your input on a few items at the end.

Quick visual overview of the `_attack()` flow after the changes (green = changed, blue = new code, orange = open question, grey = unchanged):

flow

## addressed review changes

- **C1 (degenerate regressor, `qmia_attack.py:152`):** no longer raises `RuntimeError`. Following your suggestion, it now builds a structured failed report:

  ```python
  if threshold_spread < max(1e-10, 1e-6 * score_spread):
      output = self._make_failed_output(
          target,
          "QMIA quantile regressor degenerated to a near-constant predictor ...",
      )
      try:
          self._write_report(output)
      except OSError:
          logger.warning("Could not write failed report.")
      return output
  ```

  The output carries `"status": "failed"` and `"fail_reason"`, and writes to `report.json` and `report.pdf` (a single failure-summary page that the existing `write_pdf` helper appends to any pre-existing PDF at that path, matching the convention used by other attacks). The `attack_experiment_logger` field is omitted on failure. Successful runs now also carry `"status": "success"` so consumers can check a single field.

user_stories_flow_chart drawio

- **M1 (non-finite `predict_proba`, `qmia_attack.py:118`):** same treatment as C1. No longer raises `ValueError`:

  ```python
  if not (np.isfinite(proba_train).all() and np.isfinite(proba_test).all()):
      output = self._make_failed_output(
          target,
          "target.model.predict_proba returned non-finite values; ...",
      )
      ...
      return output
  ```

- **Tests (`test_qmia_attack.py:379` and the matching M1 test):** replaced `with pytest.raises(...)` with assertions on the new output shape:

  ```python
  output = attack_obj.attack(qmia_degenerate_target)
  assert output["status"] == "failed"
  assert "degenerated to a near-constant" in output["fail_reason"]
  assert "attack_experiment_logger" not in output
  ```

  Also added a new test `test_qmia_failed_run_writes_report` that runs with `write_report=True` and confirms the JSON and PDF actually appear on disk. Also added `assert output["status"] == "success"` to `test_qmia_runs_on_binary_tabular_target` so the success status is exercised.

- **Docstrings:** added one-line docstrings to all 25 helpers across `benchmark_qmia_full.py`, `benchmark_qmia_regressor.py`, `benchmark_qmia_vs_lira.py`, and `summarize_qmia_lira_benchmark.py`.

## few small things I noticed while in the same files

- **Calibration tolerance at `qmia_attack.py:190`:** was `max(2*alpha, 0.05)` which accepted any observed FPR up to 6% for `alpha=0.01`, so the `calibration_ok` flag almost never fired. I looked into how to set a more meaningful tolerance and used the 95% binomial confidence interval half-width:

  ```python
  n_test: int = len(test_scores)
  fpr_tolerance: float = 1.96 * float(
      np.sqrt(self.alpha * (1.0 - self.alpha) / n_test)
  )
  ```

  Since `obs_fpr` is a binomial proportion. The bound shrinks with larger `n_test`, about 2% for `n=96` and 0.6% for `n=1000`. Happy to switch to a different formula if you'd prefer.

- **Wrapped `_write_report` on the failure paths in `try/except OSError`** so a disk-write error (full disk, permissions) doesn't bubble up as an exception and undo the structured-failure design:

  ```python
  try:
      self._write_report(output)
  except OSError:
      logger.warning("Could not write failed report.")
  ```

- **`_pick_fastest` / `_pick_best_auc` / `_pick_best_auc_per_sec`** in `summarize_qmia_lira_benchmark.py`: these were calling `min()`/`max()` on a filtered list without an empty-list guard, so a scenario where every attack failed would crash the summarizer with `ValueError`. They now return `None`:

  ```python
  def _pick_fastest(rows: list[dict[str, Any]]) -> dict[str, Any] | None:
      eligible = [r for r in rows if "seconds" in r]
      if not eligible:
          return None
      return min(eligible, key=lambda r: float(r["seconds"]))
  ```

  And `_print_scenario_summary` prints `"no successful runs"` for that line instead of crashing.

- **Added `plt.close()` to `_plot_lira_individuals` (`report.py:463`):**

  ```python
  plt.tight_layout()
  plt.savefig(dest)
  plt.close()
  ```

  I noticed `_roc_plot_single` and `_roc_plot` got `plt.close()` added in PR #429, and this function looked like the same pattern but was missed. Hoping that was an oversight rather than intentional, but happy to revert if you'd left it out for a reason.

## three things I'd like your input on, which I left out of scope here

### 1. `observed_public_fpr` looks like it's in-sample (`qmia_attack.py:183`)

This is the most important one; from what I can trace through the code, I noticed that the regressor is fit on `x_test_with_y` / `test_scores`, then predicted on a combined array that includes those same test rows, and `obs_fpr` is measured on that test slice:

```python
# qmia_attack.py:146 — fit on test (non-member) rows
self.quantile_model.fit(x_test_with_y, test_scores)

# qmia_attack.py:152 — predict on combined (train + the same test rows)
thresholds = self.quantile_model.predict(combined_x_with_y)

# qmia_attack.py:183 — obs_fpr measured on the test slice the regressor was fit on
obs_fpr: float = float(np.mean(y_pred_proba[len(train_scores) :, 1] > 0.0))
```

If I'm reading that right, `obs_fpr` is being measured on the same rows the quantile regressor was fit to, so I think it'll track `alpha` for any non-degenerate regressor regardless of how well the attack actually generalizes. Since this number underpins the calibration claim in the PR description, I didn't want to change how it's measured without checking with you first.

Would setting aside ~10% of `X_test/y_test` before fitting and evaluating `obs_fpr` on that held-out slice be the right approach?

### 2. Extending the "failed status run" pattern to other attacks

You mentioned this should extend to other types of attacks and right now QMIA is the only one using it. I was thinking I could lift `_make_failed_output` onto the `Attack` base class in a follow-up issue:

```python
# proposed addition to attack.py
class Attack(ABC):
    def _make_failed_output(self, target: Target, fail_reason: str) -> dict:
        """Build output dict for an attack that could not produce results."""
        ...
        return {
            "metadata": self.metadata,
            "status": "failed",
            "fail_reason": fail_reason,
        }
```

So that LiRA, WorstCase, Structural, and Attribute can adopt the same pattern.

### 3. `logging.basicConfig(level=logging.INFO)` at module import

I noticed this is used in `qmia_attack.py:30` and 9 other sacroml modules:

```python
# qmia_attack.py:30 (and 9 other modules)
logging.basicConfig(level=logging.INFO)
```

From what I've read, calling `basicConfig` at module level can sometimes affect the root logger of any application that imports the library. I wasn't sure if that's intentional in sacroml.

A proposed fix would be:

```python
# remove this line from every sacroml module
logging.basicConfig(level=logging.INFO)

# keep this line, which is the correct library pattern
logger = logging.getLogger(__name__)
```

I wanted to know whether this is worth addressing?

shamykyzer:
#### follow up: benchmark consolidation + hardening (`516d1b7`, `5fb9509`)

I had a closer look at the four scripts in `examples/sklearn/` and noticed there was a fair bit of duplicated code between them, so I merged the three benchmark scripts into one to cut things down:

- `benchmark_qmia_full.py`
- `benchmark_qmia_regressor.py`
- `benchmark_qmia_vs_lira.py`

into a single:

- `benchmark_qmia.py`

The new script keeps both `stdout` tables and `JSON` output behind a `--output stdout|json|both` flag, and `summarize_qmia_lira_benchmark.py` still works against the JSON.

While I was in there I also looked into a few small robustness fixes that I thought were worth adding:

- **`train_test_split(stratify=y)`** now falls back to a non-stratified split if the data is too small or imbalanced for stratify to satisfy.
From what I read, this is a known sklearn `ValueError` that can happen on very small or skewed inputs, so I wrapped it with a fallback rather than letting the script crash.

- **`summarize_qmia_lira_benchmark.py`** now catches `ValueError` from `_load_rows` and exits with a non-zero status. Previously it would silently exit `0` on a malformed JSON, which I thought could be misleading in CI or scripted contexts.

- **Metrics extraction** in `_run_attack` is wrapped in `try/except (KeyError, TypeError)` so a malformed attack output produces a `status` row instead of crashing.
I read through the existing failure paths (`if not output:` and `if output.get("status") == "failed":`) and adding this one extra check seemed like a sensible safety net to handle anything unexpected.

- **Failure rows** in the JSON output use a single `status` string (matching the existing `"not_attackable_or_empty"`) rather than a separate `status` + `fail_reason` pair. I went back and forth on this one but it's just a tools script for testing in the end so I shouldn't overthink it too much since it's not for the user use at the end of the day.

jim-smith:
@shamykyzer  I've set aside time later tis week to go through this new version.   as I'd like to get this into thr tooling asap and you've done loads of work on it. 
The flowchart looks super helpful, we should think about putting it into the relevant .rst files on docs so it doesn't juut get lost in this thread.

jim-smith (changes requested):
@ssrhaso All looks good to me.
**except**

Looks like your tests cover:
- all of the edge cases,  where QMIA should not work
- unit testing of individual parts 

What is missing is a test where it *should* work and we *should* see some specific records are vulnerable to attack? 

Easiest way to do that is to introduce some 'canaries' 

If you look at the (old) jupyter notebooks for the safedecisiontree() models you'll see I do something similar: take the iris data and then introduce a new record with an unseen class.

So I propose a new function 'test_qmia_predicts_canaries()' .   
In that you flip a few of the labels in the training set
 - ideally picking ones near edge of decision regions so the classifier is likely ot be affected by them. 
     - could select them using knowledge about `make_classification`, or 
     -  by getting scores for 9-nearest neighbours and looking for something with low confidence
  - keeping track of which you have flipped so you can check QMIA predicts them
  - you might want to make sure the base algorithm does over-fit to pick those up - but as you're using default settings for RandomForest it probably will.
- or failing  that, have you replicated results from the papers?


If we can prove this idea, we should be able to apply it to test other attacks

ssrhaso:
> @ssrhaso All looks good to me. **except**
> 
> Looks like your tests cover:
> 
> * all of the edge cases,  where QMIA should not work
> * unit testing of individual parts
> 
> What is missing is a test where it _should_ work and we _should_ see some specific records are vulnerable to attack?
> 
> Easiest way to do that is to introduce some 'canaries'
> 
> If you look at the (old) jupyter notebooks for the safedecisiontree() models you'll see I do something similar: take the iris data and then introduce a new record with an unseen class.
> 
> So I propose a new function 'test_qmia_predicts_canaries()' . In that you flip a few of the labels in the training set
> 
> * ideally picking ones near edge of decision regions so the classifier is likely ot be affected by them.
>   
>   * could select them using knowledge about `make_classification`, or
>   * by getting scores for 9-nearest neighbours and looking for something with low confidence
> * keeping track of which you have flipped so you can check QMIA predicts them
> * you might want to make sure the base algorithm does over-fit to pick those up - but as you're using default settings for RandomForest it probably will.
> * or failing  that, have you replicated results from the papers?
> 
> If we can prove this idea, we should be able to apply it to test other attacks

hi @jim-smith , i added test_qmia_predicts_canaries in test_qmia_attack.py following your suggestion.

the test picks 8 training rows with the lowest 9-nn same-class confidence (nearest a decision boundary), flips their labels, fits the target, then runs qmia with report_individual=True and checks that the attack flags those rows as members.

two things i hit while designing it that are worth flagging:

default randomforestclassifier (bootstrap=true) was not strong enough. each tree only sees ~63% of rows, so trees that never saw the flipped canary still vote against it and the memorisation signal gets diluted. switching to bootstrap=false makes every tree fit every row, the canary predict_proba for the flipped label hits 1.0, and the signal becomes clean. left a comment in the docstring explaining why.

ranking canaries inside the training slice (e.g. "canaries should be in the top decile of train member_probs") is bimodal across seeds. the qmia quantile threshold for (x_canary, y_flipped) has to extrapolate from test rows with that rare label/location combo, and the regressor sometimes guesses high and sometimes low. the stable metric is canary-vs-test auc, which is exactly the membership inference question applied to the rows we deliberately made vulnerable. across a sweep of seeds, dataset sizes (400-800), and class_sep (0.9-1.1), this auc was always >= 0.957.

the assertions:

output["status"] == "success"
auc of canary member_prob vs test member_prob > 0.9 (observed 0.997 on seed 0)
at least 7 of 8 canaries exceed the 90th percentile of test member_probs (observed 8/8)
full suite passes locally, 24/24 in tests/attacks/test_qmia_attack.py.

note: i did not pursue paper replication as an alternative since the canary test is what you proposed. happy to extend this once we agree the design is right, the same canary harness should drop straight onto lira / worstcase.

let me know if this is how you envisioned it, any concerns etc...

ssrhaso:
also pushed a follow-up to close out the codecov patch-coverage check that was constantly failing on this pr.

added 5 small tests in test_qmia_attack.py which are all covering the defensive branches that codecov was flagging
again, let me know if you would prefer this to be done a different way i just took initiative for now

jim-smith (approved):

jim-smith:
@ssrhaso like the canary test - that is exactly what I wanted I.e. create conditions when a half-decent attack should detect a leaky model. Good to see your tests  'independently rediscovered' the issue with bootstrap=False being much more likely to create vulnerable Random Forest's.
Suggest you (I) merge these but (ii) at the same time raise an issue to move `def test_qmia_predicts_canaries(tmp_path):` into a higher level function `def test_attack_predicts_canaries(,attack,tmp_path):`that then gets called for the QMIA, LiRA and worst case attacks

PR #436, opened 2026-04-02
chore: enable prek compatibility

### LINKED ISSUE #434: [Chore] Enable prek to run as well as pre-commit
[Prek](https://github.com/j178/prek) is a drop-in replacement for pre-commit that runs much faster. But there are some slight modifications needed here.

The `.pre-commit-config.yaml` needs to be updated to tell the spell checker to use the `pyproject.toml`:
```
  - repo: https://github.com/codespell-project/codespell
    rev: v2.4.2
    hooks:
      - id: codespell
        additional_dependencies: [tomli]
        args: ["--toml", "pyproject.toml"]
```

And then there are some JSON files in `docs/source/attacks/` that need the infinities and nans to be set to null (this aligns with #421) - I'm not quite sure why pre-commit doesn't currently pick this up yet prek does, since JSON checking is on - perhaps that could be looked at too.

ssrhaso:
Enables [prek](https://github.com/nicholasgasior/prek) to run alongside pre-commit 

Codespell config: 

Updated .pre-commit-config.yaml to pass --toml pyproject.toml to codespell (with tomli dependency), replacing the inline exclude regex. Added skip = "*.svg,*.ipynb" to [tool.codespell] in pyproject.toml to preserve the same exclusions.

JSON fixes: Replaced 11 bare Infinity values with null across docs/source/attacks/report_example_worstcase.json (10) and report_example_lira.json (1). Aligns with #421 (task I completed prior).

Why pre-commit didn't catch the JSON issue: 
Python's json.loads() accepts bare Infinity, -Infinity, and NaN as non-standard extensions. The check-json hook from pre-commit-hooks uses json.loads(), so it silently passes invalid JSON values. prek uses a stricter JSON parser that follows the spec.

Closes #434
@rpreen


rpreen (approved):
thanks, looks good

PR #440, opened 2026-04-10
feat: reduce JSON report size

### LINKED ISSUE #437: [New Feature Request] Reduce JSON report size
Currently, when running attacks that produce ROC curves (e.g., LiRA) the results written to the JSON report can be quite large. This is because arrays are included for "fpr", "tpr" and "roc_curve" which come from `metrics.py` and are later used to produce the plots that are in the PDF report. For large datasets, this can bloat the files (which are uncompressed); for example, for CIFAR10 there are 10,000 samples in the test set so the JSON file has a large number of data points for each of the above arrays, resulting in a JSON file ~1.2MB for just one attack - and this is not including the individual scores.

https://github.com/AI-SDC/SACRO-ML/blob/42041fd067daa9efe9507e26419e169768641b9a/sacroml/metrics.py#L330

We could write these particular arrays to individual files and have file paths in the JSON or perhaps the individual arrays aren't necessary since they can be computed from other arrays (e.g., ROC is computed from `y_test` and `y_pred_proba` and saving these might be more sensible/reusable).

The same probably needs to happen for the individual record scores for LiRA, which are also written to JSON.

ssrhaso:
## Summary

Reduce JSON report file size by removing large derived arrays and externalising individual record data.

- Strip `fpr`, `tpr`, and `roc_thresh` arrays from JSON serialisation via a `_json_exclude_keys` class attribute on the `Attack` base class. PDF generation is unaffected as it reads from the in-memory dict before serialisation.
- Externalise LiRA individual record scores to a compressed `.npz` file, storing a relative filename in JSON for portability. Includes `y_pred_proba` and `y_test` to allow ROC recomputation from stored data.
- Add backwards compatibility guard in `LogLogROCModule` for JSON files without ROC arrays.

For a CIFAR10-scale dataset, this reduces a single attack JSON from approximately 1.2MB to under 100KB.

Closes #437

### DIFF
--- a/sacroml/attacks/attack.py
+++ b/sacroml/attacks/attack.py
@@ -26,14 +26,17 @@ class Attack(ABC):
     # Keys stripped from JSON output for every attack. Large arrays
     # (fpr/tpr/roc_thresh) bloat reports without adding info beyond AUC; the
     # `individual` block is per-record data that we externalise to a compressed
-    # .npz file (see `_individual_npz_prefix`).
+    # .npz file (see `_npz_prefix`).
     _json_exclude_keys: frozenset[str] = frozenset(
         {"fpr", "tpr", "roc_thresh", "individual"}
     )
 
-    # Subclasses opt in to externalising per-record `individual` data to a
-    # compressed .npz file by setting a non-empty prefix (e.g. "lira").
-    _individual_npz_prefix: str = ""
+    # Subclasses opt in to externalising arrays to compressed .npz files by
+    # setting a non-empty prefix (e.g. "lira"). Two file kinds share the prefix:
+    # `<prefix>_predictions_*.npz` (always written when `_predictions_extras`
+    # returns non-empty) and `<prefix>_individual_*.npz` (written when the
+    # instance carries an `individual` block).
+    _npz_prefix: str = ""
 
     def __init__(self, output_dir: str = "outputs", write_report: bool = True) -> None:
         """Instantiate an attack.
@@ -109,14 +112,53 @@ class Attack(ABC):
         return output
 
     def _individual_extras(self, instance_key: str) -> dict[str, np.ndarray]:
-        """Return extra arrays to include in the .npz alongside `individual`.
+        """Return extra arrays to include in the per-record `individual` .npz.
 
-        Override in subclasses that need to persist arrays not part of the
-        per-record `individual` block (e.g. LiRA's `y_pred_proba`/`y_test`
-        used for ROC recomputation).
+        Override in subclasses that want to persist arrays alongside the
+        `individual` block but that aren't themselves part of it.
         """
         del instance_key
         return {}
+
+    def _predictions_extras(self, instance_key: str) -> dict[str, np.ndarray]:
+        """Return arrays needed to recompute the ROC for one instance.
+
+        Override in subclasses that produce membership predictions
+        (typically ``y_pred_proba`` and ``y_test``). Returning a non-empty
+        dict causes a ``<prefix>_predictions_<id>_<key>.npz`` to be written
+        regardless of whether per-record `individual` data is also reported,
+        so downstream tools can always reconstruct the ROC curve. The
+        default returns ``{}`` (no predictions file).
+        """
+        del instance_key
+        return {}
+
+    def _externalise_predictions(self, output: dict) -> None:
+        """Write per-instance prediction arrays to compressed .npz files.
+
+        For each instance under `attack_experiment_logger.attack_instance_logger`,
+        call `_predictions_extras` and, if it returns any arrays, write them to
+        a compressed .npz and add a `predictions_file` pointer on the instance.
+
+        This is independent of `report_individual` — the predictions file is
+        written whenever an attack supplies them, so the ROC stays recoverable
+        from disk even when per-record data is omitted from the JSON for size.
+        """
+        if not self._npz_prefix:
+            return
+        instances = output.get("attack_experiment_logger", {}).get(
+            "attack_instance_logger", {}
+        )
+        for key, inst in instances.items():
+            extras = self._predictions_extras(key)
+            if not extras:
+                continue
+            arrays: dict[str, np.ndarray] = {
+                k: np.asarray(v) for k, v in extras.items()
+            }
+            fname = f"{self._npz_prefix}_predictions_{self._instance_id[:8]}_{key}.npz"
+            np.savez_compressed(os.path.join(self.output_dir, fname), **arrays)
+            inst["predictions_file"] = fname
 
     def _externalise_individual(self, output: dict) -> None:
         """Write per-record `individual` blocks to compressed .npz files.
@@ -128,7 +170,7 @@ class Attack(ABC):
         block stays in memory (used by PDF generation and meta-attack); it is
         stripped from the JSON via `_json_exclude_keys`.
         """
-        if not self._individual_npz_prefix:
+        if not self._npz_prefix:
             return
         instances = output.get("attack_experiment_logger", {}).get(
             "attack_instance_logger", {}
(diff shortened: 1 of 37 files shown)


ssrhaso:
@rpreen Please let me know what you think of this when you get a chance!

jim-smith:
@ssrhaso does this affect the meta-attack?

ssrhaso:
> @ssrhaso does this affect the meta-attack?

@jim-smith  To my understanding and after checking thoroughly, no it does not. 

MetaAttack reads per-record scores from the in-memory sub-attack object, not from the JSON, so the key-stripping doesn't affect it.

Spotted one related side-effect while checking: the new .npz write in LIRAAttack._save_attack_metrics was gated on report_individual only, so MetaAttack sub-runs (which set write_report=False) would have dropped stray lira_individual.npz files in each sub_dir. Pushed a fix to also gate on write_report.

sacroml/attacks/likelihood_attack.py
```diff
@@ -357,6 +360,17 @@ def _save_attack_metrics(
             self.result["member"] = mia_labels
             self.attack_metrics[-1]["individual"] = self.result
 
+            if self.write_report:
+                npz_filename = "lira_individual.npz"
```
jim-smith: what if someone runs lira twice with different params? we don't want ot overwrite their individual level results so we need different filename?

jim-smith (changes requested):
why do you only remove fpr/tpr/roc_thresh for lira? they will be present in all worst-case and qmia reports as well.

Also, structural and qmia attacks should have individual records in as well, and worst-case should soon.   so the code should dealt with this.

To avoid duplication, maybe that 'stripping' code should live in the super-class so it can be called by any attack that supports individual level recording?

ssrhaso:
> why do you only remove fpr/tpr/roc_thresh for lira? they will be present in all worst-case and qmia reports as well.
> 
> Also, structural and qmia attacks should have individual records in as well, and worst-case should soon. so the code should dealt with this.
> 
> To avoid duplication, maybe that 'stripping' code should live in the super-class so it can be called by any attack that supports individual level recording?

thanks @jim-smith

on the stripping - to my understanding fpr/tpr/roc_thresh is already on the base Attack class (attack.py L25), so worst-case and structural inherit it. LiRA just extends with individual. agreed it's not obvious from the LiRA line, will add a comment.

what is LiRA-only is the .npz write which i agree should move up. I'll lift it into the base behind a class attr so structural can opt in (its individual block is currently still in the JSON), and worst-case/qmia plug in later. one thing to also note: LiRA also stores y_pred_proba and y_test in the npz for ROC recompute, which aren't part of the individual dict, so the base will need a small hook for those extras.

on the filename, i agree. I'll mint a UUID in Attack.init and use it as both the JSON log_id and the npz suffix, so two runs into the same dir don't mess up and the files stay linked.

ssrhaso:
apologies for misunderstanding initially, i believe these changes should address the stated concerns

please let me know what you think, if any mistakes or changes you recommend further
@jim-smith

jim-smith:
lgtm.
worth checking with @rpreen as it's his issue

ssrhaso:
> lgtm. worth checking with @rpreen as it's his issue

@rpreen let me know what you think, if any changes needed etc..

rpreen (dismissed):

rpreen:
At first glance this looks good - but now I am wondering if this only saves the `y_test` and `y_pred_proba` if individual scores are reported - if they aren't then it doesn't save those arrays, which are necessary to compute the ROC curve? So as it is, when `report_individual=False` it just removes the arrays from the JSON file and we have no way of computing those values?

I think `y_test` and `y_pred_proba` should always be written to their own npz files; then additionally, if `report_individual=True` then another `individual.npz` file needs to be saved that contains the individual scores, etc.

ssrhaso:
> At first glance this looks good - but now I am wondering if this only saves the `y_test` and `y_pred_proba` if individual scores are reported - if they aren't then it doesn't save those arrays, which are necessary to compute the ROC curve? So as it is, when `report_individual=False` it just removes the arrays from the JSON file and we have no way of computing those values?
> 
> I think `y_test` and `y_pred_proba` should always be written to their own npz files; then additionally, if `report_individual=True` then another `individual.npz` file needs to be saved that contains the individual scores, etc.

thanks @rpreen . pushed a fix.

the npz write is now split into two independent paths on the attack base class: a predictions file (y_test, y_pred_proba) gets written whenever the attack supplies them, regardless of report_individual, and the existing individual file still only fires when per-record data is reported. so with report_individual=False we still get lira_predictions_<id>_instance_0.npz on disk and the roc stays recomputable. added test_lira_predictions_written_without_report_individual for exactly that case.

while in there i extended the same predictions externalisation to worst-case and qmia since they were silently in the same boat. happy to scope that back to lira-only if you'd rather keep this pr tight - and open up sub issues, just thought it might be better to tackle them here whilst at it; let me know.

rpreen:
Thanks for this. However - at the risk of being a real pain - when I look at this it feels kind of ugly to have to cache these arrays inside every attack implementation. My original thinking when I created the issue was that it should be pretty simple to just write those arrays in one location, but I can see now that it's more messy because the metrics and report writing code is so intertwined (and really needs a major refactoring - and even here there is an argument to be made whether `_externalise_predictions()` and `_externalise_individual()` should live in the `Attack` class or be part of report writing). 

So, would it be simpler to just write the computed arrays that were originally in the JSON file (the fpr, tpr, etc.) since they should be able to be done in one location at report writing time and then we wouldn't need to have each attack implementation caching things, which adds to maintenance and additional work for new attacks?

ssrhaso:
> Thanks for this. However - at the risk of being a real pain - when I look at this it feels kind of ugly to have to cache these arrays inside every attack implementation. My original thinking when I created the issue was that it should be pretty simple to just write those arrays in one location, but I can see now that it's more messy because the metrics and report writing code is so intertwined (and really needs a major refactoring - and even here there is an argument to be made whether `_externalise_predictions()` and `_externalise_individual()` should live in the `Attack` class or be part of report writing).
> 
> So, would it be simpler to just write the computed arrays that were originally in the JSON file (the fpr, tpr, etc.) since they should be able to be done in one location at report writing time and then we wouldn't need to have each attack implementation caching things, which adds to maintenance and additional work for new attacks?

thanks @rpreen  i agree that the caching was the wrong approach since storing raw predictions to recompute the roc later was redundant, with no size benefit.

the new approach i implemented is what you suggested: a single generic externalisation in report.write_json. for each instance it moves the already-computed fpr/tpr/roc_thresh and the individual block into one compressed .npz sidecar, replacing them with an arrays_file pointer. it doesn't mutate the in-memory output, so pdf generation is unaffected.

as a result, all the per-attack scaffolding is gone: _npz_prefix, _predictions_extras, _individual_extras and the caches in lira/qmia/worst-case/structural. no attack caches anything now and new attacks get this for free. net change is a deletion. i also wired LogLogROCModule to read the sidecar back so the log-log roc plot still works rather than silently skipping.

happy to adjust the format or naming, or split it into separate files if you'd prefer. let me know any concerns.

ssrhaso:
@rpreen  Updated to cover Codecov

sacroml/attacks/attack.py
```diff
@@ -22,6 +22,16 @@
...
+    # Keys removed from the JSON report for every attack. The large arrays they
+    # name (ROC curves and per-record `individual` data) bloat reports without
+    # adding information beyond what the summary metrics already capture. They
+    # are externalised once, generically, to a compressed .npz sidecar at
+    # report-write time (see `report._externalise_arrays`) and replaced with an
+    # `arrays_file` pointer, so no attack needs to cache anything itself.
+    _json_exclude_keys: frozenset[str] = frozenset(
+        {"fpr", "tpr", "roc_thresh", "individual"}
+    )
+
```
rpreen, line 34: Line 190 of `sacroml/attacks/report.py` contains a similar hard-coded list that defines which arrays to write as npz - is this necessary to be defined in two locations?

ssrhaso: removed _externalise_array_keys:

 _externalise_arrays now externalises the same exclude_keys set write_json strips, flattening dict-valued keys (e.g. individual) by type instead of special-casing by name. attack._json_exclude_keys is now the single source of truth; behaviour identical, tests pass unchanged.

ssrhaso:
@rpreen Updated, please lmk

rpreen (approved):
thanks, this looks a lot cleaner

PR #441, opened 2026-04-12
feat: MetaAttack model

### LINKED ISSUE #428: [New Feature Request] Meta-attack
Currently each individual attack can calculate and output the vulnerability of each member of the training set into results.json.

However we typically report measures (accuracy, true positive rate st some fixed false positive rate, PDIF/FDIF rates)  that aggregate over all the records.

What that doesn't tell us is whether some training records are vulnerable to *all* (or most of ) the attacks.
- or to repeated runs of the same attack (since some are stochastic)
- Or whether some test records are always misclassified.

 
Something that pulled all the relevant individual vulnerabilities into a pandas data frame - one column per attack, one row per train (and possibly test) record.
- that would give us the ability to produce Venn diagrams.
- or other create a 'meta-attack' that scored 'how many different attacks is this record vulnerable to' or (since the vulnerabilities may be attack confidence values in $[0-1]$ - either the geometric or arithmetic mean of those vulnerabilities.

**Additional context**
Start by running one of the simple sklearn examples and when you run lira and structural attacks set  `report_individual=True` in the constructor
that should let you see where  the record-level vulnerabilities gt put in results .json

### LINKED ISSUE #450: Standardise per-record score field name across attacks
LiRAAttack uses `score`, QMIAAttack uses `member_prob`. Standardise the field name.

Surfaced in #441; affects LiRA, its tests, and MetaAttack's `_MIA_SCORE_FIELDS`.

### LINKED ISSUE #451: StructuralAttack ignores report_individual flag
StructuralAttack populates `record_level_results` and `attack_metrics["individual"]` unconditionally, ignoring the `report_individual` parameter (see `sacroml/attacks/structural_attack.py:480` and `:681`). Other attacks (LiRA, QMIA) respect the flag and only produce per-record output when it is True.

Surfaced during PR #441 review 

MetaAttack currently depends on this always-on behaviour to read structural scores without forcing the flag. Fixing structural to respect the flag would require MetaAttack to set `report_individual=True` for the structural sub-attack too (one-line change in `_run_sub_attack`).

### LINKED ISSUE #452: WorstCaseAttack to expose per-record scores
WorstCaseAttack outputs only aggregate metrics, no per-record scores. To include it in MetaAttack it needs a `report_individual` flag like LiRA and QMIA that populates per-record scores in `attack_metrics[N]["individual"]`.

Surfaced as Jim's PS in #441.

shamykyzer:
Closes #428

Added a **MetaAttack** class that runs multiple privacy attacks on the same target and combines their per-record results into one DataFrame.
- Runs LiRA, QMIA, and/or Structural attacks (with optional repeated runs)
- Extracts each record's vulnerability score from every attack
- Aggregates scores in two levels:
  - **Within-attack:** mean, std, and consistency across repeated runs
  - **Cross-attack:** MIA ensemble mean (arithmetic + geometric), structural flag, and a count of how many attacks flagged each record
- Outputs a `vulnerability_matrix.csv` and a standard JSON report

```python
meta = MetaAttack(
    attacks=[("lira", {"n_shadow_models": 100}, 5), ("qmia", {}), ("structural", {})],
)
meta.attack(target)
df = meta.vulnerability_df  # one row per record, one column group per attack
```

available via the factory: `factory.attack(target, "meta", attacks=[...])`


jim-smith:
@shamykyzer @ssrhaso I think this needs to be split into two different components:

1. something which can run a range of  attacks - the first of which will create report.json and the rest of which will spend to report.json (this is the current behaviour) 

2. something that, given an existing report.json, collates the results from the different attacks present (I guess, with the exception of an existing meta-attack).

The reason for wanting to separate 1 from 2 is that it will potentially save huge amounts of time- e.g. on some of the experimental results where we have  LiRa  and structural attacks for ~70K distinct combination of model-hyper parameters-dataset ,   **those  took weeks to run**

I'll go through the code base so far  to comment.

docs/specs/2026-04-12-meta-attack-design.md
```diff
@@ -0,0 +1,239 @@
...
+
+| Attack | Path after `obj.attack(target)` | Score field | Type | Length |
+|--------|-------------------------------|-------------|------|--------|
+| LiRA | `obj.attack_metrics[-1]["individual"]` | `"score"` | list[float] [0,1] | n_train + n_test |
+| QMIA | `obj.attack_metrics[0]["individual"]` | `"member_prob"` | list[float] [0,1] | n_train + n_test |
+| Structural | `obj.record_level_results` | `.k_anonymity`, `.class_disclosure`, `.smallgroup_risk` | list[int], list[bool], list[bool] | n_train only |
+
+**Record ordering:** LiRA and QMIA: train first, then test. Structural: train only.
+
+**Critical detail:** Structural always computes `record_level_results` regardless of `report_individual`. MetaAttack only needs to force `report_individual=True` on MIA attacks.
```
jim-smith: good catch. that needs fixing in structural attack

docs/specs/2026-04-12-meta-attack-design.md
```diff
@@ -0,0 +1,239 @@
...
+## Two-Level Aggregation
+
+### Level 1: Within-attack (across n_reps)
+
+For each MIA attack with n_reps > 1:
+- `{name}_mean`: mean of scores across reps, per record
+- `{name}_std`: std of scores across reps, per record
+- `{name}_consistency`: fraction of reps where score > mia_threshold, per record
+
+For structural with n_reps > 1:
```
jim-smith: structural attacks should never has n_reps >1, or if they do they should be identical because they are reporting deterministic values like k-anonymity, and whether there are sufficient residual degrees of freedom etc.

docs/specs/2026-04-12-meta-attack-design.md
```diff
@@ -0,0 +1,239 @@
...
+
+## Score Extraction Paths (verified)
+
+| Attack | Path after `obj.attack(target)` | Score field | Type | Length |
+|--------|-------------------------------|-------------|------|--------|
+| LiRA | `obj.attack_metrics[-1]["individual"]` | `"score"` | list[float] [0,1] | n_train + n_test |
+| QMIA | `obj.attack_metrics[0]["individual"]` | `"member_prob"` | list[float] [0,1] | n_train + n_test |
+| Structural | `obj.record_level_results` | `.k_anonymity`, `.class_disclosure`, `.smallgroup_risk` | list[int], list[bool], list[bool] | n_train only |
+
+**Record ordering:** LiRA and QMIA: train first, then test. Structural: train only.
```
jim-smith: good point, probably for consistency we should put in values for test items- but they would be fairly meaningless

docs/specs/2026-04-12-meta-attack-design.md
```diff
@@ -0,0 +1,239 @@
...
+
+Rows: n_train + n_test (train first, then test)
+Index: record_0, record_1, ...
+```
+
+When n_reps=1 for a MIA attack, `_std` = 0.0 and `_consistency` = 1.0 or 0.0.
+
+## Sub-Attack Isolation
+
+Each sub-attack gets its own output subdirectory:
```
jim-smith: This design decision changes one we've already made at a project level, namely that report.json should contain all the information from running various attacks on a target.
The work Hannah is doing on redesigning the report ot outputs checkers  relies on that. - and also (I think) custom reporting work other end users are doing.

docs/specs/2026-04-12-meta-attack-design.md
```diff
@@ -0,0 +1,239 @@
...
+MetaAttack computes its own global metrics using `mia_mean` as a membership predictor:
+```python
+y_pred_proba = np.column_stack([1 - mia_means, mia_means])
+membership = np.array([1]*n_train + [0]*n_test)
+self.attack_metrics = [metrics.get_metrics(y_pred_proba, membership)]
+```
+
+This gives the meta-attack its own AUC, TPR, Advantage, etc.
+
+## Report JSON Structure
```
jim-smith: This structure should exactly mirror that used by other reports.

And when a meta-attack is run win a target with an existing report.json (becuase we have already made design choices about which attacks to run - e.g. not LiRTA for huge models) the output from the meta-attack should be appended to the report.json like the other attacks do

docs/specs/2026-04-12-meta-attack-plan.md
```diff
@@ -0,0 +1,196 @@
...
+- `sacroml/attacks/factory.py` (EDIT)
+
+**Work:**
+- Create `MetaAttack(Attack)` with full `__init__`:
+  - Parse and validate `attacks` list via `_parse_attacks()`
+  - Load `k_threshold` from ACRO if None
+  - Store `mia_threshold`, `k_threshold`, `vulnerability_df`
+- Implement `_parse_attacks()`: normalise 2-tuple to 3-tuple, validate names against `SUPPORTED_ATTACKS`
+- Stub abstract methods: `attackable`, `_attack`, `_get_attack_metrics_instances`, `_make_pdf`, `__str__`
+- `attackable` fully implemented: `target.has_model() and target.has_data()`
```
jim-smith: in the way I'm suggesting, you could also check that report.json already exists I.e. you can only run the meta-attack **after** existing attacks have already been run

docs/specs/2026-04-12-meta-attack-plan.md
```diff
@@ -0,0 +1,196 @@
...
+- Add `"meta": MetaAttack` to factory registry
+- Use lazy import in factory to avoid load-time issues: `from sacroml.attacks.meta_attack import MetaAttack`
+
+**Commit:** `feat: add MetaAttack class skeleton and factory registration`
+
+**Verify:** `python -c "from sacroml.attacks.meta_attack import MetaAttack"` imports without error. `MetaAttack(attacks=[("lira", {})]).attack_specs` returns `[("lira", {}, 1)]`. Passing `"worstcase"` raises `ValueError`.
+
+---
+
+## Stage 2 — Sub-attack orchestration
```
jim-smith: This (stage 2)is the bit I am proposing you separate (or just comment) out so that:
-  we can run your meta attacks on targets where we have already run a range of attacks.
- the user has flexibility about whether to run the really time-consuming ones or not

Ultimately if we want to create new targets and run attacks on them, it is a fairly simple matter of shell scripting ...

sacroml/attacks/qmia_attack.py
jim-smith, line 1: assuming you can make the minor changes I've suggested to qmia, we can merge those into main and then pull main into here so this commit doesn't include code related to any other attacks

examples/sklearn/qmia_lira_scenarios.example.json
jim-smith, line 1: assuming you can make the minor changes I've suggested to qmia, we can merge those into main and then pull main into here so this commit doesn't include code related to any other attacks

examples/sklearn/benchmark_qmia_vs_lira.py
jim-smith, line 1: assuming you can make the minor changes I've suggested to qmia, we can merge those into main and then pull main into here so this commit doesn't include code related to any other attacks

examples/sklearn/benchmark_qmia_regressor.py
jim-smith, line 1: assuming you can make the minor changes I've suggested to qmia, we can merge those into main and then pull main into here so this commit doesn't include code related to any other attacks

examples/sklearn/benchmark_qmia_full.py
jim-smith, line 1: assuming you can make the minor changes I've suggested to qmia, we can merge those into main and then pull main into here so this commit doesn't include code related to any other attacks

sacroml/attacks/report.py
jim-smith, line 1: assuming you can make the minor changes I've suggested to qmia, we can merge those into main and then pull main into here so this commit doesn't include code related to any other attacks

sacroml/attacks/utils.py
jim-smith, line 1: assuming you can make the minor changes I've suggested to qmia, we can merge those into main and then pull main into here so this commit doesn't include code related to any other attacks

tests/attacks/test_factory.py
jim-smith, line 1: assuming you can make the minor changes I've suggested to qmia, we can merge those into main and then pull main into here so this commit doesn't include code related to any other attacks

CHANGELOG.md
jim-smith, line 1: assuming you can make the minor changes I've suggested to qmia, we can merge those into main and then pull main into here so this commit doesn't include code related to any other attacks

jim-smith: and then the changelog should be amended include the new meta-attack!

README.md
jim-smith, line 1: assuming you can make the minor changes I've suggested to qmia, we can merge those into main and then pull main into here so this commit doesn't include code related to any other attacks.

BUT the readme should include a brief reference ot the meta-attack

tests/attacks/test_qmia_attack.py
jim-smith, line 1: assuming you can make the minor changes I've suggested to qmia, we can merge those into main and then pull main into here so this commit doesn't include code related to any other attacks

tests/attacks/test_meta_attack.py
jim-smith, line 1: I'll wait until you;ve refactored before I check this code

sacroml/attacks/meta_attack.py
```diff
@@ -0,0 +1,504 @@
...
+        For each attack specification the method:
+        1. Runs the sub-attack *n_reps* times, each in an isolated subdirectory.
+        2. Extracts per-record scores from each run.
+        """
+        # {name: [[scores_rep0], [scores_rep1], ...]}  for MIA
+        # {name: [{"k_anonymity": [...], ...}, ...]}   for structural
+        mia_scores: dict[str, list[list[float]]] = {}
+        structural_scores: dict[str, list[dict]] = {}
+
+        for name, params, n_reps in self.attacks:
```
jim-smith: Easiest thing to do to achieve the refactoring I have suggested is to introduce a three-way flag 'behaviour' control entry ot this loop.

Step 1: do an initial parse of report to see what is there already -store in dictionary "already_present"

then:
``` python
if behaviour =='use_existing_only':
            for name, params, n_reps in self.attacks:
                      pass
            # skip the call to run all the  attacks
elif behaviour=="run_missing":
            for name, param, n_reps in self.attacks:
                   if name-param-n_reps not in already_present
                         run as needed
else: #default is run_all"
     your code 

** I guess you could achieve the same effect by making it a binary flag "run_missing" but that could get lengthy if you accidentally put an incorrect param

File: sacroml/attacks/meta_attack.py [hunk 9]
```diff
@@ -0,0 +1,504 @@
...
+        from sacroml.attacks.factory import create_attack  # noqa: PLC0415
+
+        sub_params = copy.deepcopy(params)
+
+        # Force per-record reporting on MIA attacks.
+        # Structural always computes record_level_results regardless.
+        if name in MetaAttack.MIA_ATTACKS:
+            sub_params["report_individual"] = True
+
+        # Isolate each run in its own subdirectory under self.output_dir.
```
jim-smith: again, if you want to leave this in then control it by a boolean 'keep_separate' or something so that the default behaviour  doesn't mess with what others  are doing I.e. each attack appends to the same report.json

sacroml/attacks/meta_attack.py
```diff
@@ -0,0 +1,504 @@
...
+        sub_dir = os.path.join(self.output_dir, f"{name}_run{run_idx}")
+        sub_params["output_dir"] = sub_dir
+        sub_params["write_report"] = False
+
+        attack_obj = create_attack(name, **sub_params)
+        result = attack_obj.attack(target)
+        if not result:
+            raise RuntimeError(
```
jim-smith: reflecting comments from the qmia code, we need a group decision about how to handle glitches.

I am not in favour of things that make it look as if the programme has crashed. Erros are for unexpected behaviour, where's I thing we should anticipate this might happen and behave gracefully without stopping a whole workflow that might include meta-attacks after a lira attack

sacroml/attacks/meta_attack.py
```diff
@@ -0,0 +1,504 @@
...
+                f"The target may not be attackable by this attack type."
+            )
+        return attack_obj
+
+    # ------------------------------------------------------------------
+    # Score extraction
+    # ------------------------------------------------------------------
+
+    _MIA_SCORE_FIELDS: dict[str, str] = {
+        "lira": "score",
```
jim-smith: Good catch. 
maybe we should rename the LuRA attack field to member_prob. 
We should be consistent between attacks.

That could be a job for  separate issue though as we've already got results using the name 'score'

sacroml/attacks/meta_attack.py
```diff
@@ -0,0 +1,504 @@
...
+        # LiRA stores metrics as a list; QMIA also uses a list.
+        # Both place the "individual" dict in attack_metrics[N].
+        for metrics_dict in attack_obj.attack_metrics:
+            if "individual" in metrics_dict:
+                scores = metrics_dict["individual"][field]
+                # Clip to [0, 1]: default offline mode is already bounded,
+                # but Carlini modes can produce unbounded log-likelihood ratios.
+                return [max(0.0, min(1.0, s)) for s in scores]
+
+        raise RuntimeError(
```
jim-smith: see earlier comment about making it look as if the programme has crashed

sacroml/attacks/meta_attack.py
```diff
@@ -0,0 +1,504 @@
...
+            f"{name} attack did not produce individual scores. "
+            f"Ensure report_individual=True was set."
+        )
+
+    @staticmethod
+    def _extract_structural_scores(attack_obj: Attack) -> dict:
+        """Return per-record structural risk indicators.
+
+        Reads directly from the ``record_level_results`` dataclass, which
+        is always populated regardless of ``report_individual``.
```
jim-smith: there should be some things which open at the level of the target e.g. is the _minimum_ k_anonymity too small,

and some that happen at record level e.g. what is the k-anonymity of _this_ record? 

we could (and the code used  to) map all the target-level values onto individual levels values by simply repeating them but it took up lots of space

jim-smith (commented):
## This is all good stuff

Hopefully I've pointed out where your code can be split into two (or have to decision paths) without too much hassle.

The rationale is  to make the available as a general purpose attack that **_can_ if desired_** orchestra missing attacks, but that can reuse the computational effort of attacks people have already run.

- so it should cope with new attacks we haven't written yet (hence the need ot standardise names e.g. `score` vs `member_prob`

- and it should be able to cope with case where people have chosen not to run LiRA, or maybe only run it with a few shadow models

### Ps we need to change the worst-case attack code to output individual scores

jim-smith:
@shamykyzer Tagging you in case you didn't see the comments above.   Not trying to rush you to respond :)

shamykyzer:
> @shamykyzer Tagging you in case you didn't see the comments above. Not trying to rush you to respond :)

Hi @jim-smith, will do, next week I'll dedicate more time on both this and the instance-based attack. 

Thanks!

docs/specs/2026-04-12-meta-attack-design.md
shamykyzer: Filed as #451 (sub-issue of #428) to keep this tracked. Note that MetaAttack currently relies on the always-on `record_level_results` to read structural scores, so when StructuralAttack starts respecting `report_individual`, MetaAttack's `_run_sub_attack` will need a one-line change to set `report_individual=True` for the structural sub-attack too.

shamykyzer:
Hi @jim-smith, could you please review these changes that I addressed:

- 3-way `behaviour` flag (`run_all`, `use_existing_only`, `fill_missing`) per your sketch
- MetaAttack output appended to `report_dir/report.json` by default (`keep_separate=True` opts out)
- `use_existing_only` and `fill_missing` read both canonical single-file `report.json` and subdirectory layouts
- `_make_pdf` produces a PDF via `report.create_meta_report` with a bar chart of records flagged by N attacks
- CHANGELOG and README mention MetaAttack
- Graceful degradation across sub-attacks (narrow except, warnings on expected failures)
- Structural `n_reps` clamped to 1 (deterministic)

Follow-ups (separate PRs / issues):
- `score` vs `member_prob` standardisation: #450
- StructuralAttack ignores `report_individual` flag: #451
- WorstCaseAttack to expose per-record scores: #452
- Coverage gap (currently 87%)

Thanks.

---

**Overview**
meta_attack_overview

**Architecture flow with behaviour modes**
meta_attack_architecture

**Aggregation pipeline**
meta_attack_aggregation

<details>
<summary><b>Class hierarchy</b> (click to expand)</summary>

meta_attack_class_hierarchy

</details>

shamykyzer:
also -- thank you Jim for the thorough review and the time you put into walking through the code.

examples/sklearn/meta_attack_example.py
```diff
@@ -0,0 +1,103 @@
...
+        X_train_orig=X_train,
+        y_train_orig=y_train,
+        X_test_orig=X_test,
+        y_test_orig=y_test,
+    )
+    for idx in range(X.shape[1]):
+        target.add_feature(f"feature_{idx}", [idx], "float")
+
+    # --- Run MetaAttack ---
+    meta = MetaAttack(
```
jim-smith, line 56: As this is a example it would be a good idea  to provide a value for the parameter that says whether to use existing, create missing, or run all?

sacroml/attacks/meta_attack.py
```diff
@@ -0,0 +1,828 @@
...
+    "qmia": "member_prob",
+}
+"""Maps factory key → field name inside ``attack_metrics[N]["individual"]``.
+
+Used only by :meth:`MetaAttack._extract_mia_scores` (the live-attack path).
+The disk-reading path (:meth:`MetaAttack._extract_scores_from_report`) uses
+the same field names but looks them up directly rather than via this mapping.
+"""
+
+_EPS: float = 1e-10
```
jim-smith: Suggest we create a file `constants.py` then define it _EPS in  there and import from there e.g.  a similar EPS is defined (with  a different value) in attacks/utils.py

sacroml/attacks/meta_attack.py
```diff
@@ -0,0 +1,828 @@
...
+        base class.  The CSV (``vulnerability_matrix.csv``) and PDF always
+        follow the JSON output location.
+    """
+
+    def __init__(
+        self,
+        attacks: list[tuple | list],
+        behaviour: str = "run_all",
+        report_dir: str | None = None,
+        mia_threshold: float = 0.5,
```
jim-smith: can we change this to `mia_threshold: float = DEFAULT_MIA_THRESHOLD,` and define `DEFAULT_MIA_THRESHOLD` either at the top of the file or (better) in a new file `attacks/constsnts.py`

sacroml/attacks/meta_attack.py
```diff
@@ -0,0 +1,828 @@
...
+        k_threshold: int | None = None,
+        output_dir: str = "outputs",
+        write_report: bool = True,
+        keep_separate: bool = False,
+    ) -> None:
+        super().__init__(output_dir=output_dir, write_report=write_report)
+        # MetaAttack does not use shadow models; remove the empty directory
+        # created by the base class so the output directory stays clean.
+        with contextlib.suppress(OSError):
```
jim-smith, line 131: just checking, this doesn't remove shadow models created by other attacks ? that would be catastrophic@!

jim-smith:
@shamykyzer have made one comment in the example file. Will make a local copt of this branch and see what it gives me when I run it.

examples/sklearn/meta_attack_example.py
shamykyzer: good idea, thanks, added `behaviour="run_all"` to the example with a comment listing the two alternatives so the three modes show up side by side

sacroml/attacks/meta_attack.py
shamykyzer: good thing to check, that would be catastrophic if true, as i read it the safety holds via two lines, but please gut check me

base class at `attacks/attack.py:43`:
```python
self.shadow_path = os.path.normpath(f"{self.output_dir}/shadow_models")
```
so `self.shadow_path` is rooted at this metaattack instance's own `output_dir`, it can never resolve to another attack's directory

then at `meta_attack.py` lines 131 to 134:
```python
with contextlib.suppress(OSError):
    os.rmdir(self.shadow_path)
```
`os.rmdir` raises `OSError` on a populated directory which `suppress` swallows, so the call does nothing rather than destructively deleting if anything is in there

i think the worst case is "remove the empty `shadow_models/` this metaattack just created, otherwise do nothing"

ssrhaso:
added unit tests covering the previously-uncovered branches in meta_attack.py (defensive guards, report-scanning edge cases, sub-attack failure paths, non-numeric scores) and the finite-auc branch in create_meta_report. 

takes meta_attack.py from 87% to 100% and should clear the codecov/patch and codecov/project checks. test-only, no changes to attack code :) 

@shamykyzer

shamykyzer:
hi @jim-smith, just wanted to let you know this PR is ready for review

the constants file is in, the `behaviour` kwarg is in the example and README, and hasaan added the coverage on the 22nd

the follow ups are each their own PR now, #460 structural respecting `report_individual`, #461 worstcase per record scores, #462 canary test parametrised over QMIA and LiRA (worstcase is a follow up), #463 lira score to member_prob rename

mind taking another look when you have a chance

jim-smith (approved):

PR #446, opened 2026-04-30
feat: sklearn-compatible scorers for mia metrics (groundwork for #350)

### LINKED ISSUE #350: [New Feature Request] Improve the "worst case" attack
Currently the the worst case attack takes an attack classifier and a single combination of hyperparameters to train the attack model - some form of autotuning would seem necessary - at the moment the user would do this by creating multiple worst case attack objects; potentially a range of models or hyperparameters could be passed to the attack and a tuning framework like [optuna](https://optuna.org) could be used internally.

ssrhaso:
## what

adds sklearn-compatible scorers wrapping `sacroml.metrics.get_metrics` so mia metrics can be passed to `gridsearchcv` / `randomizedsearchcv` via the standard `(estimator, X, y)` protocol.

## files

- `sacroml/attacks/_scorers.py`: `SCORERS` dict (auc, tpr@0.1, tpr@0.001, fdif01, fdif02, advantage, pdif01) + `resolve_scorer()` helper that also passes sklearn scoring strings through unchanged.
- `tests/attacks/test_scorers.py`: 7 tests covering protocol, value parity with `get_metrics`, passthrough, and error paths.

## why

groundwork for #350. `worstcaseattack` is not modified in this pr. follow-up pr will add `attack_model_param_grid` and `tuning_metric` and use these scorers as the search `scoring=` argument.

## status

draft, opened to review the scorer surface before wiring into the attack.

### DIFF
--- a/sacroml/attacks/worst_case_attack.py
+++ b/sacroml/attacks/worst_case_attack.py
@@ -144,7 +144,11 @@ class WorstCaseAttack(Attack):
             attack_model_param_grid
         )
         self.search_type: str = search_type
+        if not isinstance(search_n_iter, int) or search_n_iter < 1:
+            msg = f"search_n_iter must be a positive integer; got {search_n_iter!r}."
+            raise ValueError(msg)
         self.search_n_iter: int = search_n_iter
+        resolve_scorer(tuning_metric)
         self.tuning_metric: str | Callable = tuning_metric
         self.dummy_attack_metrics: list = []
         self._tuned_params: dict | None = None
@@ -313,6 +317,19 @@ class WorstCaseAttack(Attack):
         # instantiate attack model
         return model(**params) if params is not None else model()
 
+    def _make_rep_splitter(self, random_state: int) -> StratifiedShuffleSplit:
+        """Splitter used by both the tuning search and the rep loop.
+
+        Both call sites must produce byte-identical folds for the
+        "tuning counts toward n_reps" property to hold; centralising
+        construction here prevents the two from drifting.
+        """
+        return StratifiedShuffleSplit(
+            n_splits=self.n_reps,
+            test_size=self.test_prop,
+            random_state=random_state,
+        )
+
     def _resolve_param_grid(self) -> dict | list[dict] | None:
         """Return the parameter grid, or ``None`` if tuning is disabled."""
         grid = self.attack_model_param_grid
@@ -355,11 +372,7 @@ class WorstCaseAttack(Attack):
                 f"at least 2 folds; got n_reps={self.n_reps}."
             )
             raise ValueError(msg)
-        splitter = StratifiedShuffleSplit(
-            n_splits=self.n_reps,
-            test_size=self.test_prop,
-            random_state=random_state,
-        )
+        splitter = self._make_rep_splitter(random_state)
         scorer = resolve_scorer(self.tuning_metric)
         base_estimator = self._get_attack_model()
         if self.search_type == "grid":
@@ -407,9 +420,15 @@ class WorstCaseAttack(Attack):
                 strict=True,
             )
         ]
+        best_idx = int(np.flatnonzero(cv_results["rank_test_score"] == 1)[0])
+        best_candidate_per_fold_scores = [
+            float(cv_results[f"split{i}_test_score"][best_idx])
+            for i in range(self.n_reps)
+        ]
         self._tuning_info = {
             "best_params": dict(search.best_params_),
             "best_score": float(search.best_score_),
+            "best_candidate_per_fold_scores": best_candidate_per_fold_scores,
             "tuning_metric": (
                 self.tuning_metric
                 if isinstance(self.tuning_metric, str)
@@ -485,11 +504,7 @@ class WorstCaseAttack(Attack):
         if self._tuned_params is not None:
             # Use the same CV splits the search ran over so the "reps" and
             # the "tuning CV folds" coincide (the tuning counts toward n_reps).
-            splitter = StratifiedShuffleSplit(
-                n_splits=self.n_reps,
-                test_size=self.test_prop,
-                random_state=split[0],
-            )
+            splitter = self._make_rep_splitter(split[0])
             fold_splits = list(splitter.split(mi_x, mi_y))
         else:
             # Legacy path: variable random states per rep (one resample each).
--- a/sacroml/attacks/worst_case_attack.py
+++ b/sacroml/attacks/worst_case_attack.py
@@ -3,17 +3,23 @@
 from __future__ import annotations
 
 import logging
-from collections.abc import Iterable
+from collections.abc import Callable, Iterable
 from typing import Any
 
 import numpy as np
 from fpdf import FPDF
 from sklearn.base import BaseEstimator
 from sklearn.metrics import confusion_matrix
-from sklearn.model_selection import train_test_split
+from sklearn.model_selection import (
+    GridSearchCV,
+    RandomizedSearchCV,
+    StratifiedShuffleSplit,
+    train_test_split,
+)
 
 from sacroml import metrics
 from sacroml.attacks import report
+from sacroml.attacks._scorers import resolve_scorer
 from sacroml.attacks.attack import Attack
 from sacroml.attacks.target import Target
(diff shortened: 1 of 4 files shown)


jim-smith (approved):

ssrhaso:
overview of what i've added on top of the scaffold.

**what changed**

`worstcaseattack` now accepts an optional sklearn style parameter grid. when supplied, the attack runs a `gridsearchcv` or `randomizedsearchcv` over the membership inference data before the rep loop. the search picks one set of hyperparameters using the configured scoring metric, and those tuned parameters drive every subsequent rep (both real and dummy). when no grid is supplied, the legacy path is preserved exactly.

**how the tuning is plumbed**

following @jim-smith 's directive that tuning counts toward `n_reps`, the search cross validation and the rep loop share a single `stratifiedshufflesplit(n_splits=n_reps, test_size=test_prop, random_state=reproduce_split[0])`. because that splitter is deterministic given the random_state, the search folds and the rep folds are byte identical, so the tuning is not paid for on top of the reps. the splitter mirrors the existing per rep `train_test_split` semantics (repeated stratified shuffle splits rather than disjoint n way folds), so each rep still sees a fresh resample. jim flagged the slight conceptual gap between repeated shuffle splits and a true n way cv partition; this implementation accepts that small difference rather than changing the rep semantics underneath users.

**new init params**

* `attack_model_param_grid` (`dict | list[dict] | "default" | none`, default `none`): the search space. `"default"` resolves to a built in grid for `randomforestclassifier`, `mlpclassifier`, or `logisticregression`. other models require an explicit grid.
* `search_type` (`"grid" | "random"`, default `"grid"`).
* `search_n_iter` (`int`, default `10`): only used by the randomised search.
* `tuning_metric` (`str | callable`, default `"auc"`): any key in `sacroml.attacks._scorers.scorers`, any sklearn scoring string, or a callable on the `(estimator, x, y)` protocol.

**reporting**

a new `metadata["tuning"]` block carries `best_params`, `best_score`, `tuning_metric`, `search_type`, `param_grid`, `n_candidates`, and a per candidate `cv_results` list with `params`, `mean_test_score`, `std_test_score`, and `rank_test_score`. the rest of the report (`attack_experiment_logger`, dummy block, global metrics) is unchanged. when no grid is configured, `metadata["tuning"]` is absent.

**backward compatibility**

with `attack_model_param_grid=none`, the rep loop still uses `train_test_split` with the same per rep random_state derived from `reproduce_split`. splits and metrics are byte identical to the previous behaviour and the eight pre existing worst case tests pass unchanged.

**limitations and possible follow ups**

* `search_cv` is not exposed. decoupling cv folds from `n_reps` would break the "tuning counts toward n_reps" property and reintroduce the "tuning paid on top of reps" cost, so it felt like a footgun. if a future use case calls for cheaper tuning on top of many reps, adding a `search_cv` override and reverting to the legacy rep loop when set is straightforward.
* the search uses `n_jobs=1`. parallel tuning would help on large grids but interacts with reproducibility and ci, left as a follow up.
* dummy attacks reuse the tuned hyperparameters rather than retuning per dummy. this preserves apples to apples comparison; per dummy retuning is a small extension if reviewers prefer it.
* `_default_param_grids` covers only the three classifiers named in the thread. extending to others (e.g. `gradientboostingclassifier`) is one line per model.
* callable `tuning_metric` values are stored as `__name__` (or `repr`) in `metadata["tuning"]`, since the live callable is not json serialisable. a richer scorer registry form could replace this if useful for the metric rank meta experiments.
* the metric rank experiment from §4.1 of the tdp paper is intentionally absent here; per @jim-smith it belongs in sacro research. 

**tests**

eight pre existing worst case tests pass unchanged. eight new tests cover the default grid path, a custom dict grid with dummy reuse, randomised search with the `tpr@0.1` scorer, `n_reps < 2` rejection, unknown model default rejection, invalid grid string, invalid `search_type`, and the untuned metadata staying clean. the fourteen scorer tests continue to pass. one unrelated factory test fails only when sequenced after the lira/qmia suites in a full suite run; in isolation it passes.

happy to split the scorer commit and the tuning commit into two prs if that is easier to review, or if you have any other concerns which you would like me to directly address - just wasnt sure how much further i should proceed.

ssrhaso:
did some small follow-ups on top of the tuning commit:

- factored the `StratifiedShuffleSplit` construction in `_tune_if_needed` and `run_attack_reps` into a single `_make_rep_splitter` helper. i believe both sites must produce byte-identical folds for the "tuning counts toward n_reps" property to hold; centralising prevents drift.

- `metadata["tuning"]` now carries `best_candidate_per_fold_scores` (winner's `split{i}_test_score` from `cv_results_`), so per-fold tuning stability is visible without rerunning.

- `search_n_iter < 1` and an unknown `tuning_metric` now fail in `__init__` rather than at fit time.

- two new tests cover the validation paths; existing default-grid test extended to check the per-fold field. all 32 tests pass.

JessUWE:
@jim-smith

sacroml/attacks/worst_case_attack.py
```diff
@@ -95,7 +140,19 @@ def __init__(
...
         self.attack_model_params: dict[str, object] | None = attack_model_params
+        self.attack_model_param_grid: dict | list[dict] | str | None = (
+            attack_model_param_grid
+        )
+        self.search_type: str = search_type
+        if not isinstance(search_n_iter, int) or search_n_iter < 1:
+            msg = f"search_n_iter must be a positive integer; got {search_n_iter!r}."
+            raise ValueError(msg)
+        self.search_n_iter: int = search_n_iter
+        resolve_scorer(tuning_metric)
```
jim-smith: you're not catching the return of `resolve_scorer(tuning_metric)`

Shouldn't this be 
```
try:
     resolved_tuning_metric= resolve_scorer(tuning_metric)
     self.tuning_metric:str|Callable= resolved_tuning_metric
except ValueError:
    ...print message set to to some default ...
```

sacroml/attacks/worst_case_attack.py
jim-smith: rather than leave this as hard-coded within the function (never, ever, a good idea) it would be better to define a dict DEFAULT_RF_PARAMS at the top of the file and then set `params=DEFAULT_RF_PARAMS` here .

sacroml/attacks/worst_case_attack.py
```diff
@@ -255,6 +317,134 @@ def _get_attack_model(self) -> BaseEstimator:
         # instantiate attack model
         return model(**params) if params is not None else model()
```
jim-smith, line 333: worth splitting this line into and/or adding comment to make it clear you only reach the final option if the user has asked for a model of type that is not random forest, AND has not provide parameters.e.g.
`#resort to default options for other model types with no prams provided`

jim-smith (changes requested):
3 things to address is worst_case_attack.py please

sacroml/attacks/worst_case_attack.py
ssrhaso: __init__ now wraps resolve_scorer in try/except, caches the resolved scorer on self._resolved_tuning_scorer, and falls back to "AUC" with a logger.warning on ValueError. _tune_if_needed reads the cached scorer instead of re-resolving, so resolution happens exactly once per instance.

sacroml/attacks/worst_case_attack.py
ssrhaso: defined _DEFAULT_RF_PARAMS at module scope next to _DEFAULT_PARAM_GRIDS and _get_attack_model now reads from it. apologies for not catching this :)

sacroml/attacks/worst_case_attack.py
ssrhaso: reordered the conjunction so the params is None check reads first, and added a comment above the return noting the fallthrough is only reached when the attack model is not random forest and the user did not supply attack_model_params. let me know if you think it's too excessive

ssrhaso:
> 3 things to address is worst_case_attack.py please



> 3 things to address is worst_case_attack.py please

i think i have addressed all of your concerns, please let me know what you think moving forward

ssrhaso:
@jim-smith Please let me know what you think

jim-smith (approved):

PR #453, opened 2026-05-15
feat(pytorch): capture batch_size as optional Target parameter

### LINKED ISSUE #447: [New Feature Request] Set PyTorch batch size appropriately
Currently the [PyTorch wrapper](https://github.com/AI-SDC/SACRO-ML/blob/main/sacroml/attacks/model_pytorch.py) sets the batch size to 32, but it should instead use the same as the original training because it has a large influence on performance. 

I think the reason 32 was originally set was since the hardware used to train the model might differ from the hardware used to assess it - if it was trained on a machine with a giant GPU it may not run at all on a smaller one. A possible solution to that might be to use linear scaling of the learning rates proportional to the batch size as this seems to be used in practice to achieve roughly comparable nets (bigger batch size = bigger learning rate) - but if done then this should be shown clearly as a warning to the person running something like LiRA or perhaps presented as an option.

A complicating factor is that the `batch_size` parameter is not currently supplied in the `train_params` because it has been considered part of the dataset handler, for example:

https://github.com/AI-SDC/SACRO-ML/blob/a7ebc8e141556d62064839200fa9c504e48ef7d4/examples/pytorch/simple/dataset.py#L61-L70

and that function is used in the `Target` class when loading the dataset initially, but then undergoes the conversion from dataloader to numpy as explained in #411 

https://github.com/AI-SDC/SACRO-ML/blob/a7ebc8e141556d62064839200fa9c504e48ef7d4/sacroml/attacks/target.py#L191-L197

A possible solution might be if #411 was addressed and the train/test indices for the different shadow models were passed to the `get_dataloader()` function in the supplied dataset module - but that seems like a big change. 

**For now, if might be simpler to capture the `batch_size` explicitly as an optional parameter in the `Target` constructor (and update the examples)  and then use that in the `PytorchModel` class.**

ssrhaso:
closes #447

## what

the pytorch wrapper hardcoded `batch_size=32` for (re)training and inference. batch size strongly affects model behaviour, so shadow models at 32 may not match a target trained differently, distorting attack results (e.g. lira). this captures `batch_size` as an optional `Target` parameter and uses it in `PytorchModel` (richard's suggested "for now" approach).

## changes

- `PytorchModel`: optional `batch_size` (default 32) on `__init__`/`load()`, used in `fit()`, `predict()`, `predict_proba()`, propagated via `clone()`.
- `fit()`: optional `batch_size` override, warns on mismatch with recorded value.
- `Target`: optional `batch_size` field, persisted in `target.yaml` as a top-level scalar (not in `train_params`).
- examples (simple + cifar): batch size made explicit, passed to both dataloader and `Target`.
- tests: round-trip, default-to-32, old-yaml back-compat, fit override warning.

## notes

- `predict`/`predict_proba` now use `self.batch_size` instead of 32. outputs unchanged (eval mode, batchnorm running stats); only memory/throughput differs.
- batch size is user-supplied, not auto-detected: the eager dataloader to numpy conversion (the still-open part of #411) leaves no batch info to recover. defaults to 32 when unset.
- default-to-32 warning fires once at load time in `_load_model()`, not per `_wrap_model()` call (avoids per-shadow-model log spam).

old targets without `batch_size` load fine, default to 32, warn once.

## out of scope

- linear lr scaling for train/assess hardware mismatch: deferred to a follow-up issue. i think it is slightly out of scope for here and deserves its own issue

## testing

57 tests pass locally (`test_pytorch`, `test_lira_attack`, `test_attacks_target`, `test_qmia_attack`, `test_worst_case_attack`, `test_factory`) plus 4 new batch_size tests.

let me know what you think, any changes etc.. @rpreen

### DIFF
--- a/sacroml/attacks/target.py
+++ b/sacroml/attacks/target.py
@@ -80,6 +80,11 @@ class Target:
         Path to module containing training function.
     train_params : dict or None
         Hyperparameters for training the model.
+    batch_size : int or None
+        Batch size used to train the original target model. Threaded to
+        PyTorch model (re)training and inference. Should match the original
+        training batch size since it materially affects attack model fidelity
+        (e.g. shadow models). When None, PyTorch defaults to 32.
     dataset_name : str
         The name of the dataset.
     dataset_module_path : str
@@ -122,6 +127,7 @@ class Target:
     model_params: dict | None = None
     train_module_path: str = ""
     train_params: dict | None = None
+    batch_size: int | None = None
 
     # Dataset attributes
     dataset_name: str = ""
@@ -165,6 +171,10 @@ class Target:
                 train_params=self.train_params,
             )
         if isinstance(model, torch.nn.Module):
+            # Resolve None -> 32 silently here: _wrap_model() runs per Target
+            # construction (and once per shadow model in LiRA), so warning here
+            # would spam the log. The unspecified-batch-size warning instead
+            # fires exactly once at load time in _load_model().
             return PytorchModel(
                 model=model,
                 model_path=self.model_path,
@@ -173,6 +183,7 @@ class Target:
                 model_params=self.model_params,
                 train_module_path=self.train_module_path,
                 train_params=self.train_params,
+                batch_size=self.batch_size if self.batch_size is not None else 32,
             )
         if isinstance(model, (SklearnModel, PytorchModel)):
             return model
@@ -373,6 +384,12 @@ class Target:
             }
         )
 
+        # batch_size is Target-level metadata, persisted as a top-level scalar
+        # (not nested in train_params, which is forwarded verbatim as kwargs to
+        # the user's train() function).
+        if isinstance(self.model, PytorchModel):
+            target["batch_size"] = self.model.batch_size
+
         # Copy module files
         if self.model_module_path:
             shutil.copy2(self.model_module_path, os.path.join(path, "model.py"))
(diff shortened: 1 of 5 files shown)


sacroml/attacks/target.py
```diff
@@ -165,6 +171,10 @@ def _wrap_model(self, model: ModelLike) -> WrappedModel:
                 train_params=self.train_params,
             )
         if isinstance(model, torch.nn.Module):
+            # Resolve None -> 32 silently here: _wrap_model() runs per Target
+            # construction (and once per shadow model in LiRA), so warning here
+            # would spam the log. The unspecified-batch-size warning instead
+            # fires exactly once at load time in _load_model().
```
rpreen: I appreciate this comment from a review perspective, but I don't think it needs to go into the merged code; line 186 is pretty self-explanatory.

sacroml/attacks/target.py
```diff
@@ -373,6 +384,12 @@ def _save_model(self, path: str, ext: str, target: dict) -> None:
             }
         )
 
+        # batch_size is Target-level metadata, persisted as a top-level scalar
+        # (not nested in train_params, which is forwarded verbatim as kwargs to
+        # the user's train() function).
```
rpreen: I think this comment should be removed

sacroml/attacks/target.py
```diff
@@ -422,14 +439,36 @@ def _load_model(self, path: str, target: dict) -> None:
...
+                path, target.get("train_module_path", "")
+            ),
+            "train_params": target.get("train_params", {}),
+        }
+
+        # batch_size only applies to PyTorch models. _load_model() is called
+        # exactly once per Target.load(), so emitting the warning here (rather
+        # than in _wrap_model(), which runs per shadow model) guarantees it
+        # fires exactly once for an old target.yaml lacking the key. No dedupe
+        # flag or warnings filter is needed.
```
rpreen: Again, I think the code is self-explantory

rpreen (changes requested):
This looks ok - I just think a couple of the comments should be removed to keep the source clean

rpreen (approved):
thanks

PR #458, opened 2026-05-21
refactor: move unwrap_model to sacroml.attacks.utils for reuse

### LINKED ISSUE #455: refactor: move _unwrap_model to utils.py
@jim-smith suggested on #431 ([thread](https://github.com/AI-SDC/SACRO-ML/pull/431#discussion_r3138185029)) that `_unwrap_model` in `instance_based_attack.py` would be useful to other attacks and should move to `sacroml/attacks/utils.py`

Tracking here, will open a separate PR to move it and update the import

shamykyzer:
Closes #455

- Lifts `_unwrap_model` out of `InstanceBasedAttack` and into `sacroml.attacks.utils` as `unwrap_model` so other attacks can reuse it, per Jim's suggestion
- Added four direct tests in `tests/attacks/test_utils.py`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ Changes:
     membership thresholds. A sample is predicted as a member when its observed score
     exceeds the predicted threshold at quantile level (1 - alpha). No shadow models or
     architecture knowledge required. Registered in the attack factory as `"qmia"`.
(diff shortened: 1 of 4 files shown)


shamykyzer:
Folded this into PR #431 instead so the InstanceBasedAttack work lands as one PR, since the unwrap_model refactor only touches code that's in #431 anyway

PR #460, opened 2026-05-25
fix(structural): gate per-record results behind report_individual flag

### LINKED ISSUE #428: [New Feature Request] Meta-attack

### LINKED ISSUE #451: StructuralAttack ignores report_individual flag

shamykyzer:
closes #451

- `StructuralAttack` accepts a `report_individual` flag in its constructor, but the `record_level_results` dataclass and the `attack_metrics["individual"]` dict were populated regardless of the flag
- the existing check inside `_get_attack_metrics_instances` (line 696) was effectively dead because `record_level_results` was always truthy by the time it ran
- i wrapped the dataclass construction (line 480 to 484) and the `attack_metrics["individual"]` write (line 681) in `if self.report_individual:`
- i initialised `self.record_level_results` to `None` in `__init__` so the attribute exists before `_attack()` runs, mirroring how `LIRAAttack` gates its `result` dict
- `MetaAttack` continues to work because `meta_attack.py:530` already sets `sub_params["report_individual"] = True` for every attack it spawns, so the consumer path stays populated
- this branch is stacked on `428-meta-attack` until that PR lands, so the diff here will shrink once that merges

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@ Changes:
     membership thresholds. A sample is predicted as a member when its observed score
     exceeds the predicted threshold at quantile level (1 - alpha). No shadow models or
     architecture knowledge required. Registered in the attack factory as `"qmia"`.
+*   Fix: `StructuralAttack` now respects the `report_individual` flag. Per-record
+    `record_level_results` and `attack_metrics["individual"]` are only populated when the
+    flag is set to `True`, matching the behaviour of `LIRAAttack` and `QMIAAttack`.
 
 ## Version 1.4.3 (Jan 29, 2026)
 
--- a/sacroml/attacks/structural_attack.py
+++ b/sacroml/attacks/structural_attack.py
@@ -360,6 +360,7 @@ class StructuralAttack(Attack):
         super().__init__(output_dir=output_dir, write_report=write_report)
         self.target: Target | None = None
         self.results: StructuralAttackResults | None = None
+        self.record_level_results: StructuralRecordLevelResults | None = None
         self.report_individual = report_individual
 
         # Load risk appetite from ACRO config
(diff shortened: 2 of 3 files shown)


jim-smith (approved):
@shamykyzer sorted, thanks

PR #467, opened 2026-06-19
Add sharey option to subplots in_plot_lira_individuals report.py

### LINKED ISSUE #466: [Behaviour Change] tweak plotting for lira
**Is your  request related to a problem? Please describe.**
When lira runs it produces two plots showing the distribution of lira probabilities for the train and test sets.
Annoyingly the y-axes are set automatically and independently so they don't always align even though the scale should always be 0-1

**Describe the alternative behaviour you'd like**

Change the way the subplots re produced so they shared a common y-axis

N/A
**Additional context**
Fixed by the extra parameter in the code below
```
def _plot_lira_individuals(metrics: dict, dest: str) -> None:
    """Create a plot of the individual record LiRA scores."""
    scores = np.array(metrics["individual"]["score"])
    member = np.array(metrics["individual"]["member"])

    _, axes = plt.subplots(1, 2, figsize=(12.4, 4.8),sharey=True)
...
```

jim-smith:
Fixes minor irritation

### DIFF
--- a/sacroml/attacks/report.py
+++ b/sacroml/attacks/report.py
@@ -552,7 +552,7 @@ def _plot_lira_individuals(metrics: dict, dest: str) -> None:
     member = np.array(metrics["individual"]["member"])
 
     _, axes = plt.subplots(1, 2, figsize=(12.4, 4.8), sharey=True)
-
+    axes[0].set_ylim(ymin=-0.05, ymax=1.05)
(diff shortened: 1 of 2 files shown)


jim-smith:
@rpreen nudge

JessUWE (approved):
minimal change. lgtm

Issue #414, opened 2026-02-25
[New Feature Request] Enable attacks to work with regression models


rpreen:
Currently attacks only work on classification models, it shouldn't be too much work to make them work for regression.

rpreen:
The attribute attack exploits having knowledge of the correct label, so this might need to be changed to some kind of distance measure.

PR #472, opened 2026-07-10
fix: report generalisation gap instead of inverted accuracy gap

### LINKED ISSUE #468: [Behaviour Change] Generalisation error metric is mislabelled
`StructuralAttack` reports the "generalisation error" but it is actually reporting the accuracy gap between the train and test set (it is also inverted so returns a negative value). Generalisation error is conventionally understood to be the test set error.

shamykyzer:
As reported in #468, `StructuralAttack`'s generalisation metric was inverted and mislabelled: it returned the train/test accuracy gap with the sign flipped, so an overfit model showed a *negative* value.

This changes it to report the **generalisation gap** = test error - train error (for classification, `(1 - test_acc) - (1 - train_acc)`), which is **positive when the model overfits**. Classification only; regression is left to #414 as discussed on the issue.

- Flip the sign in the `SklearnModel` and `PytorchModel` generalisation methods.
- Clarify the docstrings, and fix the swapped `train_acc`/`test_acc` descriptions in `StructuralAttackResults`.
- Update the affected test assertions; add a CHANGELOG entry.

Closes #468

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -13,7 +13,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v6
+        uses: actions/checkout@v7
 
       - name: Build
         run: pipx run build
--- a/.github/workflows/sphinx-docs.yml
+++ b/.github/workflows/sphinx-docs.yml
@@ -16,7 +16,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v6
+        uses: actions/checkout@v7
 
       - name: Install Pandoc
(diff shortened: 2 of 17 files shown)


rpreen:
Thanks for this. Please can you remove any references to issue numbers to keep the source clean? Also the variable/function names (pretty much anywhere that deals with "generalisation error") like `generalisation_error` need to be renamed to generalisation gap for clarity.

rpreen (changes requested):
This looks ok, but `sacroml/config/target.py‎` seems to include unrelated changes that need to be removed first?

rpreen (approved):

PR #473, opened 2026-07-10
feat(meta): add struct_vuln_rule option for combining structural risk flags

### LINKED ISSUE #464: [New Feature Request] meta attack should have options for constructing overall vulnerability of structural attack
`meta_attack.py` computes an overall vulnerability from structural results:

```
           data["struct_vuln"] = [
                (k < self.k_threshold or cd or sg)
                for k, cd, sg in zip(k_vals, cd_vals, sg_vals, strict=True)
            ] + none_pad
```
it should be an option whether ot run this an a logical OR or AND 
maybe this logic should be in `structural_attack.puy`
**Additional context**

shamykyzer:
closes #464

- `struct_vuln` in the meta attack vulnerability matrix was hard coded to flag a record when any structural indicator fired (k below threshold, class disclosure, or smallgroup risk)
- this adds a `struct_vuln_rule` option to `MetaAttack` accepting `"or"` (default, behaviour unchanged) or `"and"` so the flag can require all indicators instead, as the issue asks
- i moved the combination itself into `structural_attack.combine_risk_flags()` since as i read the issue that felt like where it belongs, next to the code that produces the indicators, happy to move it back into meta_attack.py if you'd rather
- the chosen rule shows up in `attack_params` and `global_metrics` in report.json so downstream readers can tell which rule produced the flags
- the rule is validated at construction time like `behaviour`, so a typo fails before any sub attack runs rather than mid workflow

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,12 @@ Changes:
     (multi-section, as produced when individual attacks append to the same file)
     and any subdirectory-per-attack layout. Registered in the attack factory as
     `"meta"`.
+*   Feat: `MetaAttack`: new `struct_vuln_rule` option selecting how per-record
+    structural indicators (k-anonymity below threshold, class disclosure,
+    small-group risk) combine into the `struct_vuln` flag: `'or'` (default,
+    unchanged behaviour) or `'and'`. Combination logic moved to
+    `structural_attack.combine_risk_flags()`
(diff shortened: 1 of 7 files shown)


JessUWE:
This PR has been opened for quite a while and  it would be good to have it reviewed, the conflict merge issue has been fixed

PR #480, opened 2026-08-19
fix: metrics report TPR at percentage FPRs by default

### LINKED ISSUE #475: [Behaviour Change] Default `fpr_perc=True` to match Carlini et al. (2022) reporting convention
Currently the calculations for TPR at low FPRs defaults to not using percentages, which would be a 100x looser metric and not what is expected since following Carlini et al. (2022) the most reported values are TPR @ 0.1% FPR and TPR @ 0.001% FPR.

Shouldn't the default `fpr_perc` be set to True?

https://github.com/AI-SDC/SACRO-ML/blob/b7885f1bb4481ffbe2df1a5e0f056a2fddc73650/sacroml/metrics.py#L38-L73

If so, it might be clearer to also add "%" to the end of the metric name:

https://github.com/AI-SDC/SACRO-ML/blob/b7885f1bb4481ffbe2df1a5e0f056a2fddc73650/sacroml/metrics.py#L325-L328

JessUWE:
This PR updates TPR-at-FPR metrics to use percentage-based FPR values by default, matching the reporting convention used by Carlini et al. (2022).

Metric names now include the percentage sign, for example:

•⁠  ⁠⁠ TPR@10% ⁠
•⁠  ⁠⁠ TPR@0.1% ⁠

The related scorer keys, report metrics, tuning configuration, and tests have been updated accordingly.

### DIFF
--- a/sacroml/attacks/report.py
+++ b/sacroml/attacks/report.py
@@ -20,10 +20,10 @@ DISPLAY_METRICS = [
     "Advantage",
     "FDIF01",
     "PDIF01",
-    "TPR@0.1",
-    "TPR@0.01",
-    "TPR@0.001",
-    "TPR@1e-05",
+    "TPR@10%",
+    "TPR@1%",
+    "TPR@0.1%",
+    "TPR@0.001%",
 ]
 
 MAPPINGS = {"PDIF01": lambda x: np.exp(-x)}
--- a/sacroml/attacks/report.py
+++ b/sacroml/attacks/report.py
@@ -20,10 +20,10 @@ DISPLAY_METRICS = [
     "Advantage",
     "FDIF01",
     "PDIF01",
-    "TPR@10%",
-    "TPR@1%",
     "TPR@0.1%",
+    "TPR@0.01%",
     "TPR@0.001%",
+    "TPR@1e-05%",
 ]
 
 MAPPINGS = {"PDIF01": lambda x: np.exp(-x)}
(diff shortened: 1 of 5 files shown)


sacroml/metrics.py
```diff
@@ -320,11 +322,11 @@ def get_metrics(
     # Add some things useful for debugging / filtering
     metrics["pred_prob_var"] = y_pred_proba.var()
 
-    # TPR at various FPR
-    fpr_vals = [0.5, 0.2, 0.1, 0.01, 0.001, 0.00001]
+    # TPR at various FPR (percentages, following Carlini et al. 2022)
+    fpr_vals = [50, 20, 10, 1, 0.1, 0.001]
```
rpreen: These values need to stay at the originals: 0.5, 0.2, etc.

tests/attacks/test_worst_case_attack.py
```diff
@@ -268,13 +268,13 @@ def test_tuning_random_search_with_alt_scorer(common_setup):
         attack_model_param_grid=grid,
         search_type="random",
         search_n_iter=3,
-        tuning_metric="TPR@0.1",
+        tuning_metric="TPR@10%",
```
rpreen: 10% would be too high - should leave at 0.1%

sacroml/attacks/report.py
```diff
@@ -20,10 +20,10 @@
...
     "FDIF01",
     "PDIF01",
-    "TPR@0.1",
-    "TPR@0.01",
-    "TPR@0.001",
-    "TPR@1e-05",
+    "TPR@10%",
+    "TPR@1%",
+    "TPR@0.1%",
+    "TPR@0.001%",
```
rpreen: Values should stay the same but be % so 0.1 now becomes 0.1%, etc.

rpreen (approved):
lgtm

PR #484, opened 2026-09-23
feat: Enable attacks to work with regression models

### LINKED ISSUE #414: [New Feature Request] Enable attacks to work with regression models

JessUWE:


jim-smith:
@JessUWE @rpreen assume you are liaising on this?

